PR #6, opened 2015-11-07
Readme updates

AnthonySteele:
better wordings, fixes to code samples and added section headings

open questions:
- what does "IStatsDPublisher.MarkEvent" do?

- In the context of an IoC container, can `IStatsDPublisher` be a singleton, must it be transient or some other lifestyle?

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ This syntax for timers less typing in simple cases, where you always want to tim
    }
 ```
  
-The `StartTimer` returns an `IDisposable` that wraps a stopwatch. The stopwatch is automatically stopped and the metric sent when it falls out of scope on the closing `}` of the using statement.
+The `StartTimer` returns an `IDisposable` that wraps a stopwatch. The stopwatch is automatically stopped and the metric sent when it falls out of scope on the closing `}` of the `using` statement.
  
 ```csharp
    //  timing a lambda without a return value:
--- a/README.md
+++ b/README.md
@@ -4,9 +4,7 @@
 
 ### TL;DR
 
-We use this within our components to publish [statsd](http://github.com/etsy/statsd) metrics from .Net code.
-
-We've been using this in most of our things, since around 2013.
+We use this within our components to publish [statsd](http://github.com/etsy/statsd) metrics from .NET code. We've been using this in most of our things, since around 2013.
 
 ### Features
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,30 +4,28 @@
...
 * statsd metrics formatter
 * UDP client handling
 
-####
+#### Publishing statistics
+
+`IStatsDPublisher` is the interface that you will use in most circumstances. With this you can `Increment` or `Decrement` an event, and send values for a `Gauge` or `Timing`.
 
-`IStatsDPublisher` is the interface that you will use in most circumstances. Use an instance of this in order to send events, timers and gauges.
+The concrete class that implements `IStatsDPublisher` is `StatsDImmediatePublisher`. For the constructor parameters, you will need the statsd server host name. You can change the standard port (8125). You can also append a prefix to all stats. These values often come from configuration as the host name and/or prefix may vary between test and production environments.
```
petemounce: "append a prefix" sounds wrong. Prepend. We should add the ability to have both prefix and suffix, what with the difficulty we have with inconsistently named metrics...

README.md
```diff
@@ -36,25 +34,26 @@ example of Ioc in NInject for statsd publisher with values from config:
...
+		stats.Increment("DoSomething-Attempt");
 		var stopWatch = Stopwatch.StartNew();
-
         var success = DoSomething();
 
 		stopWatch.Stop();
 		if (success)
         {
-			stats.Timing("DoSomething.Success", stopWatch.Elapsed);
+			stats.Timing("DoSomething-Success", stopWatch.Elapsed);
```
petemounce: I'd keep the `.` here. Segments with more than one piece of logical data are less flexible.

README.md
```diff
@@ -4,30 +4,28 @@
 
 ### TL;DR
 
-We use this within our components to publish [statsd](http://github.com/etsy/statsd) metrics.
+We use this within our components to publish [statsd](http://github.com/etsy/statsd) metrics from .Net code.
```
petemounce: nit: `.NET`.

README.md
AnthonySteele: Good spot, I have fixed it.

If you want to add another config option for a stat suffix then you would need another constructor parameter.I'm not a great fan of injecting primitive types such as these `string` and `int` using an IoC container. It's a less-used feature of the various containers, meaning that support and syntax is going to vary between different containers; developer familiarity will be lower etc. 

So when you find yourself doing more of it, there may be a better way. i.e. making a type to hold all these values, so that programmer and IoC container have a unique type to work with. It also allows for sensible defaults.
e.g. 
```csharp
public class StatsDSettings
{
   public int Port { get; set; }
   public string Host { get; set; }
  // etc
}

public class StatsDPublisher
{
   public StatsDPublisher(StatsDSettings settings)
   {
   // etc
  }
}

```

AnthonySteele:
Regarding future work on more configuration options for prefixes ans suffixes:

If you want to add another config option for a stat suffix then you would need another constructor parameter.I'm not a great fan of injecting primitive types such as these `string` and `int` using an IoC container. It's a less-used feature of the various containers, meaning that support and syntax is going to vary between different containers; developer familiarity will be lower etc. 

So when you find yourself doing more of it, rather than "add more parameters to the method", there may be a better way. i.e. making a type to hold all these values, so that programmer and IoC container have a unique type to work with. It also allows for sensible defaults.
e.g. 
```csharp
public class StatsDSettings
{
   public int Port { get; set; }
   public string Host { get; set; }
  // etc
}

public class StatsDPublisher
{
   public StatsDPublisher(StatsDSettings settings)
   {
   // etc
  }
}

```

AnthonySteele:
Can we merge this? It may not be perfect, but I think it's an improvement.

petemounce:
Yes, totally happy to iterate!

PR #8, opened 2015-12-05
Renamable simple timers

AnthonySteele:
Introduce a new interface, `IDisposableTimer` to allow a simple timer in use to be renamed after it has been started.

It looks like:
````csharp
public interface IDisposableTimer : IDisposable
{
    string StatName { get; set; }
}
```

so you can, e.g.

````csharp
  using (var timer = publisher.StartTimer("statName."))
  {
      var success = DoSomething();
      timer.StatName += success ? "Ok" : "Fail";
   }
```
We don't want the stat name to be empty or null at the end of the using block. As it stands it must have a value at the start, but can be set to null without error, as this isn't a problem unless it stays that way. So the behaviour is to throw an exception if the name is null or empty at the end.


Open question before merge :

- Is `StatName` a good name for the property? I prefer it to `Bucket` but neither seems perfect


src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -5,7 +5,7 @@ namespace JustEat.StatsD
 {
     public static class TimerExtensions
     {
-        public static IDisposable StartTimer(this IStatsDPublisher publisher, string bucket)
+        public static IDisposableTimer StartTimer(this IStatsDPublisher publisher, string bucket = null)
```
tonyharverson: This feels like a stealthy way to cancel a stat, which could lead to suprising behaviour (lost into the aether) if the dev forgets to initialise a bucket.  It'd probably be better to either a) insist on a non-null-or-empty value for the bucket here (and in the setter), or b) raise an exception if the bucket is null at the end.  If anyone absolutely feels the need for a cancel() use case, we can add it explicitly later.

AnthonySteele: So the idea behind starting with a null/empty string is to allow for the case of "I am going to set the stat name in entirety later (e.g. depending on operation success or failure), so there's no point in putting anything in now".

However you are right, it does make the default case an unwelcome one - this does not encourage the dev to "fall into the pit of success" where the stat has a name.

So I will
- raise an exception if the bucket is null at the end
- possibly also revert this line and insist on a string to start off with. The dev can still specify null explicitly at start, but this is more visible.

petemounce:
I think we would improve the usability of the library if we introduced methods that took an instance representing a metric, that can then be formatted to a statsd v1 string representation, or other representations like logstash JSON or metrics v2. That would greatly help metrics being consistently named and discoverable.

AnthonySteele:
That's probably a separate concern. I'm not familiar at all with metrics v2.

brainmurphy:
Looks good to me. Build is green: https://ci.appveyor.com/project/justeattech/justeat-statsd/build/1.0.4.47

PR #10, opened 2016-02-26
Target .NET 4.5.0

GraemeF:
While Microsoft no longer supports .NET 4.5.0, Mono doesn't yet support everything from .NET 4.5.2 so targeting .NET 4.5.2 means this library can't be used on Mono.

This PR changes the target of the JustEat.StatsD project to target .NET 4.5.0, thus allowing use on Mono.


AnthonySteele:
I'm not thrilled by this, but it may be necessary.

PR #12, opened 2016-08-11
Support .NET Core

martincostello:
Support .NET Core by adjusting the `UdpClient` constructor used and using `Dns.GetGetHostAddressesAsync()` instead.

Requires a further commit to update the `nuspec` to correctly declare the package as supporting .NET Core before being merged.


src/JustEat.StatsD/EndpointLookups/DnsEndpointProvider.cs
```diff
@@ -10,7 +10,8 @@ public class DnsEndpointProvider : IDnsEndpointMapper
     {
         public IPEndPoint GetIPEndPoint(string hostName, int port)
         {
-            return new IPEndPoint(Dns.GetHostAddresses(hostName)[0], port);
+            var endpoints = Dns.GetHostAddressesAsync(hostName).Result;
```
AnthonySteele: Are we ok with the use of `.Result`, or do we have to make the whole thing async?

martincostello: I thought about that, but to properly make it all async means either breaking changes to add new methods to the interfacesand/or change the signatures.

Alternatively we could add new interfaces that support async and implement them where appropriate?

AnthonySteele: Probably not worth it ... yet

martincostello:
Looks like the .NET Portability Analyzer didn't find everything, so there's some more changes required.

martincostello:
Well, it builds. The dependencies for .NETStandard1.6 might need reducing, and the AppVeyor build doesn't append the build number to the package version yet.

The biggest job left appears to be sorting the tests out.

src/JustEat.StatsD/StatsDMessageFormatter.cs
```diff
@@ -5,7 +5,9 @@
 
 namespace JustEat.StatsD
 {
+#if NETDESKTOP
```
stuart-lang: You could use the standard ```NET45``` & ```NETSTANDARD1_6``` compilation symbols rather than define your own. Or a lot of people define their own per usage, so here would be something like ```#if BINARY_SERIALATION``` and then in the project.json you can mix and match what you need per TFM, although this arguably has less value now that we have netstandard.

martincostello: Interesting - I wasn't aware there were built in ones. That's much simpler.

src/JustEat.StatsD/project.json
```diff
@@ -0,0 +1,68 @@
...
+      }
+    },
+    "netstandard1.6": {
+      "buildOptions": {
+        "define": [ "NETCORE" ]
+      },
+      "dependencies": {
+        "NETStandard.Library": "1.6.0",
+        "System.Net.NameResolution": "4.0.0"
+      }
```
stuart-lang: The dependencies can be flattened and reduced to:
```
"System.Diagnostics.Tools": "4.0.1",
"System.Net.NameResolution": "4.0.0",
"System.Net.Sockets": "4.1.0",
"System.Runtime.InteropServices": "4.1.0"
```

src/JustEat.StatsD/project.json
```diff
@@ -0,0 +1,68 @@
...
+
+  "frameworks": {
+    "net45": {
+      "buildOptions": {
+        "define": [ "NETDESKTOP" ]
+      },
+      "frameworkAssemblies": {
+      }
+    },
+    "netstandard1.6": {
```
stuart-lang: You can target netstandard1.3 with no additional changes

stuart-lang:
Loving your work 👍
Me & @JosephWoodward will hopefully soon be coming up to FPH on a friday to hack on some dotnet core things in the afternoon, including JustSaying with @AnthonySteele, wanna join us? I've yet to sort out the details.

martincostello:
@stuart-lang Yeah sure, that sounds good! I've resolved all your feedback. Just a Travis CI build and fixing the tests left now I think.

stuart-lang:
@martincostello are you still waiting on the travis build, or are you happy with this?

martincostello:
@stuart-lang As well as a Travis build to verify it works there, also the tests need reworking still as the dependencies don't support .NET Core and I wasn't sure what to do about that as it's a non-trivial task.

martincostello:
FYI I've had issues with this code running in .NET Core, so there's some more discussion needed about exactly what we want to do to support it. In my case in a separate fork I switched to using ```UdpClient.ctor()``` and sending the packets using ```SendTo(byte[], IPEndpoint)```.

stuart-lang:
@martincostello I'll try to help. So what happens when you call it?

martincostello:
A ```SocketException``` is thrown (socket error 10049 I think). The constructor seems to suggest a local endpoint is required, not a remote one, which is why it throws. The ```SendTo()``` method wants a remote one, and that works fine. I can show you the fork version tomorrow.

slang25:
Cool. Is it newer than this: https://github.com/martincostello/JustEat.StatsD/tree/Support-DotNet-Core

martincostello:
Yeah, I've effectively forked it internally within the consumer at the moment. I'll switch back to this once we get the PR how we want it and merged.

slang25:
Cool, I'll take a look tomorrow 👍

RTodorov:
Hello guys! Do you still plan to merge this?

martincostello:
@RTodorov Hi, we're still to agree exactly how to migrate the test suite as well as changes to APIs used for .NET Core support. Once that's agreed plus I've got time to fix up the PR with the agreed changes we'll hopefully get it merged in short-order. Sorry it's been taking so long.

martincostello:
I've done some work fixing this up today, but the test dependencies are a bit of a mess due to the dependency on JustBehave.  I tried just targeting .NET Framework 4.5.1 in the tests, but the NuGet dependencies didn't seem to want to restore.

Any opinions on the best way to proceed? For example, we could re-write the tests to just be vanilla NUnit tests without the need for JustBehave and its dependency chain.

src/JustEat.StatsD/SimpleObjectPool.cs
```diff
@@ -8,7 +8,7 @@ namespace JustEat.StatsD
     public sealed class SimpleObjectPool<T>
         where T : class
     {
-        private readonly Stack<T> _pool;
+        private readonly ConcurrentStack<T> _pool;
```
slang25: This is better. Although I'm not sure why it was a stack in the first place. I would have thought a ```ConcurrentBag``` would be a better fit.
If you look on MSDN for object pooling, ```ConcurrentBag``` is what they use:
https://msdn.microsoft.com/en-us/library/ff458671(v=vs.110).aspx

martincostello: Good point. Updated as suggested in 556fa7743bbcff8d474a9688c4e8cbdffe44e68a.

slang25:
Regarding the test project, just bump it to net452 (from net451) for now, then you'll be able to restore all the packages. I'd like to separate out the AutoFixture dependency in JustBehave, or move that piece out into a separate nuget, that way we could move these without a big refactor.

slang25:
Here's a set of dependencies that work for the test project:
```
  "dependencies": {
    "JustEat.StatsD": "1.1.0-*",
    "AutoFixture": "3.50.2",
    "AutoFixture.AutoFakeItEasy": "3.50.2",
    "FakeItEasy": "1.25.3",
    "JustBehave": "1.0.0.12",
    "NLog": "5.0.0-beta03",
    "NUnit": "3.5.0",
    "Shouldly": "2.6.0",
    "dotnet-test-nunit": "3.4.0-beta-3"
  },
```
There are then a couple of minor code changes to make to get them to run and pass.

martincostello:
Thanks @slang25! Tests restored in c386c6b8e8756ae8247f0df5851f2f9bb69d5b98 and running successfully in the AppVeyor CI from 03bb34fd489ebd970fc99e217030be97090ffa47 with .NET Framework 4.5.2.

I think now we just need to try out the [latest package from AppVeyor](https://ci.appveyor.com/project/justeattech/justeat-statsd/build/1.1.0.73/artifacts) in a .NET Framework and .NET Core feature and check that it works OK functionally, then we should be good to merge.

martincostello:
Tests OK with .NET Core, just need to investigate an error on the full Framework with:

```
iisexpress.exe Error: 0 : General Exception when sending metric data to statsD :- Message : One or more errors occurred., Inner Exception System.Net.Sockets.SocketException (0x80004005): No such host is known
   at System.Net.Dns.HostResolutionEndHelper(IAsyncResult asyncResult)
   at System.Net.Dns.EndGetHostAddresses(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization), StackTrace    at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at JustEat.StatsD.EndpointLookups.DnsEndpointProvider.GetIPEndPoint(String hostName, Int32 port)
   at JustEat.StatsD.StatsDUdpClient.Send(IEnumerable`1 metrics).
```

martincostello:
That error was actually caused by a copy-paste configuration error on my part in the end. **_Sigh_**

Metrics successfully being published to StatsD from both an application targeting ,NET Core and .NET Framework 4.5.2 using ```1.1.0-beta-0073```.

slang25:
Nice, shall I merge, then we can get it on nuget.org as a beta?

martincostello:
Yeah, pushing a beta version to NuGet.org for some wider pre-release testing would be good before we go all-in and push ```1.1.0```.

martincostello:
@RTodorov We've published a beta of the [NuGet package](https://www.nuget.org/packages/JustEat.StatsD/1.1.0-beta0075) that supports .NET Core if you'd like to try it out.

RTodorov:
Thanks @martincostello, I'll definitely check it out! Thanks for the good work!

PR #13, opened 2016-09-13
Functional style with access to the timer 

AnthonySteele:
Functional style with access to the timer for setting the timer name
There are a lot of function type signatures here: 4 methods to support 2 yes/no options.
Which is not ideal but it's not that hard to use.

So the idea is that you can then do for the example with all the options:

``` csharp

_statds.Time("myHttpApi.", async timer => {
   var response = await _myHttpApi.GetSomeData(id);
   timer.StatName += (int)response.StatusCode;
   return response;
});
```

Http is the canonical case for having access to the stat name, as we typically want to time http requests, and group these times by the response code, but we don't know the response code until more than halfway through the code being timed.


martincostello:
Do you plan an doing anything with this. Shall we close it? Shall we merge it?

AnthonySteele:
I want to keep it, but if we're making breaking changes we can also remove some overloads, so don't merge for now.

AnthonySteele:
I have updated this PR, it is mergeable. No conflicts with recent work- Like with the disposable timers, it is convenient syntax for putting a timer around something, but doesn't really interact with the core of this lib much.

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -5,42 +5,72 @@ namespace JustEat.StatsD
...
             {
-                return func();
+                return func(timer);
             }
         }
 
-        public static async Task<T> Time<T>(this IStatsDPublisher publisher, string bucket, Func<Task<T>> func)
+
+        /// <summary>
+        /// functional style for timing a lambda with return value, is async
```
slang25: Minor detail: Func / delegate != lambda. I suggest changing it to say delegate or function.

slang25:
LGTM 👍

PR #16, opened 2017-03-07
Add TCP Client

timvw74:
Adding a TCP client allows us to use this within AWS Lambdas as UDP is not supported.


src/JustEat.StatsD/StatsDTcpClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            return Send(new[] { metric });
+        }
+
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "This is one of the rare cases where eating exceptions is OK")]
+        public bool Send(IEnumerable<string> metrics)
+        {
+            var data = EventArgsPool.Pop();
+            //firehose alert! -- keep it moving!
+            if (null == data)
```
AnthonySteele: Backwards conditions, you do not need in C#

martincostello: I suspect the majority of this code was copied from the UDP implementation.

src/JustEat.StatsD/StatsDTcpClient.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        /// <summary>	Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources. </summary>
+        /// <param name="disposing">	true if resources should be disposed, false if not. </param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                _disposed = true;
+            }
+        }
```
AnthonySteele: Do we need all this for disposal? the method is essentially empty. can you Delete it or reduce it to `{ }`. 
Also the comments "Performs application-defined tasks..." are vague boilerplate, remove them

src/JustEat.StatsD/StatsDTcpClient.cs
timvw74: Martin may not be wrong.

src/JustEat.StatsD/IStatsDTcpClient.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace JustEat.StatsD
+{
+    public interface IStatsDTcpClient : IStatsDClient
+    {
+    }
```
AnthonySteele: Why?

src/JustEat.StatsD/StatsDTcpClient.cs
AnthonySteele: It's not good there either.

src/JustEat.StatsD/IStatsDTcpClient.cs
timvw74: The IStatsDClient (old IStatsDUdpClient) was public, therefore I wanted to stay consistent in case someone had taken a reference to it.

AnthonySteele: The way I see it is that we would now have two working implementations of `IStatsDClient` - a name that says nothing about the transport. Another interface with no methods adds nothing.

AnthonySteele: I see that we have `IStatsDUdpClient` before :(

src/JustEat.StatsD/StatsDClient.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        /// <summary>	Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources. </summary>
+        public void Dispose()
+        {
+            if (!_disposed)
+            {
+                _disposed = true;
+                GC.SuppressFinalize(this);
+            }
+        }
```
AnthonySteele: What's the point of this method? It doesn't actually dispose of any owned resources, just does some book-keeping of a boolean flag.

AnthonySteele:
It seems that there are some style issues that have not much to do with the task of adding Tcp Support to the statsd client. Do we want them in a different PR and are we allowing breaking changes? If so, I suggest we get away from `IStatsdUdpClient` as an interface name, the consumer should not care which transport is in use and we should not have to create empty interface subclasses.

alastairstevens:
Tim and I were having this exact conversation and decided we did not want to introduce bresaking changes whilst not being comfortable with a lot of the code.

AnthonySteele:
Fair enough, breaking changes should go in a later PR. But do you need to introduce an `IStatsdTcpClient` interface with no methods, when a breaking change after is likely to remove it?

alastairstevens:
I agree.  We will remove it

src/JustEat.StatsD.Tests/JustEat.StatsD.Tests.xproj
```diff
@@ -7,7 +7,7 @@
   <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />
   <PropertyGroup Label="Globals">
     <ProjectGuid>8d9aa1a3-2619-4800-aaec-84ae4da15455</ProjectGuid>
-    <RootNamespace>JustEat.StatsD</RootNamespace>
+    <RootNamespace>JustEat.StatsD.Tests</RootNamespace>
```
martincostello: I prefer to put the root namespace of the tests the same as the root of the thing under test, otherwise once you drop down a namespace in the tests you end up having to import namespaces all the time rather than having a one-to-one match.

Target: ```Foo.Bar``` -> ```Foo.Bar.Baz``` 
Tests : ```Foo.Bar``` -> ```Foo.Bar.Baz```

compared to:

Target: ```Foo.Bar``` -> ```Foo.Bar.Baz``` 
Tests: ```Foo.Bar.Tests``` (```Foo.Bar``` is implied) -> ```Foo.Bar.Tests.Baz``` and ```Foo.Bar.Baz```

src/JustEat.StatsD/IStatsDUdpClient.cs
```diff
@@ -1,11 +1,6 @@
-﻿using System;
-using System.Collections.Generic;
-
-namespace JustEat.StatsD
+﻿namespace JustEat.StatsD
 {
-    public interface IStatsDUdpClient : IDisposable
+    public interface IStatsDUdpClient : IStatsDClient
```
martincostello: This is still a breaking change because the inheritance hierarchy is being changed. In-place updates of the assembly wouldn't work, even if re-compiling the consuming code after updating still works.

src/JustEat.StatsD/StatsDClient.cs
martincostello: This ```IDispose``` implementation is wrong for a base class. It doesn't allow the derived class to easily override it.

It should be:

```csharp
~StatsDClient()
{
    Dispose(false);
}

public void Dispose()
{
    Dispose(true);
    GC.SupressFinalize(this);
}

protected virtual void Dispose(bool disposing)
{
    // No implementation if just an extension point for derived classes
}
```

src/JustEat.StatsD/StatsDImmediatePublisher.cs
```diff
@@ -10,80 +10,88 @@ public class StatsDImmediatePublisher : IStatsDPublisher
     {
         private static readonly CultureInfo SafeDefaultCulture = new CultureInfo(StatsDMessageFormatter.SafeDefaultIsoCultureId);
         private readonly StatsDMessageFormatter _formatter;
-        private readonly IStatsDUdpClient _transport;
+        private readonly IStatsDClient _client;
         private bool _disposed;
 
-        public StatsDImmediatePublisher(CultureInfo cultureInfo, string hostNameOrAddress, int port = 8125, string prefix = "")
+        public StatsDImmediatePublisher(CultureInfo cultureInfo, string hostNameOrAddress, int port = 8125, string prefix = "", Transport transport = Transport.Udp)
```
martincostello: New optional parameter is also a breaking change.

src/JustEat.StatsD/StatsDImmediatePublisher.cs
```diff
@@ -10,80 +10,88 @@ public class StatsDImmediatePublisher : IStatsDPublisher
...
         {
             _formatter = new StatsDMessageFormatter(cultureInfo, prefix);
-            _transport = new StatsDUdpClient(hostNameOrAddress, port);
+            if (transport == Transport.Tcp)
```
martincostello: Maybe just inject the implementation instead? Then if we add a 3rd transport this class doesn't have to change.

src/JustEat.StatsD/StatsDImmediatePublisher.cs
```diff
@@ -102,9 +110,9 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
```
martincostello: This would override ```protected virtual void Dispose(bool disposing)``` once implemented in the base class.

src/JustEat.StatsD/StatsDTcpClient.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public StatsDTcpClient(int endpointCacheDuration, string hostNameOrAddress, int port) : base(endpointCacheDuration, hostNameOrAddress, port)
+        {
+        }
+
+        public override void CallClient(SocketAsyncEventArgs data)
+        {
+            using (var tcpClient = GetTcpClient())
+            {
+                tcpClient.Client.Connect(data.RemoteEndPoint);
```
martincostello: This is going to blow up if the client creation fails as ```GetTcpClient()``` will return ```null```.

martincostello:
There's a lot of breaking changes here, so I'd suggest we publish ```1.1.0``` before merging this, even if we want to do _more_ breaking changes after it. The version number also needs bumping to ```2.0.0``` if we're breaking things.

AnthonySteele:
I suggest that we get this merged with minimum disruption.
I get that it breaks binary compatibility and so should require a version bump, but client code should be able to get the new package, rebuild and be good to go.

I have other changes in mind to interface name, constructor signature which would require clients to change the code and then rebuild.

src/JustEat.StatsD/StatsDClient.cs
slang25: I personally wouldn't bother with the dispose pattern at all (i.e. finalizer with flag). Do we want to cater for people subtyping this class with unmanaged resources? I genuinely cannot see this ever happening, because there isn't much internal state that is interesting, if you did ever want to extend this with unmanaged resources (still don't see it happening), you'd delegate to a private instance this class.

Without it the code is simpler and doesn't survive the extra GC that is unnecessary.

AnthonySteele: Agreed. It doesn't look as if anything is actually being disposed, it's just an empty template. Let's not do it at all.

timvw74:
@AnthonySteele I could remove the defult and add in a constructor that uses: ` public StatsDImmediatePublisher(CultureInfo cultureInfo, string hostNameOrAddress, int port = 8125, string prefix = "")`  that would then call the new one with UDP

PR #17, opened 2017-03-08
Update to .NET Core SDK 1.0.1 and csproj files

martincostello:
* Update to .NET Core SDK v1.0.1 and ```.csproj files``` instead of ```project.json```.
  * Update .NET Core TFM to ```netstandard1.6```.
  * **TODO**: NUnit tests don't get discovered anymore. Need to find what the new way to get it working with the .NET Core Test SDK is (or switch to xunit 2.2.0).
  * .NET Core is now only installed into the solution directory if it isn't already installed on the machine.


martincostello:
Looks like the references are a bit messed up.

martincostello:
This is the error in Visual Studio 2017 vis-a-vis NUnit:

```
MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "C:\Users\{user}\.nuget\packages\dotnet-test-nunit\3.4.0-beta-3\lib\net451\dotnet-test-nunit.exe", "AMD64". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.	
```

JosephWoodward:
I'd be inclined to suggest switching to xunit, it always seems ahead of Nunit in terms of keeping up with .NET Core.

martincostello:
Yeah I'd agree with that too. NUnit seems to lag behind by months.

JosephWoodward:
Yeah, when .NET Core first landed it took them months until it was working, where as xunit was working from the get-go. I also prefer xunit's API over nunit, but that's just preference.

martincostello:
What's your opinion @slang25 and @AnthonySteele ?

AnthonySteele:
I'm much more familiar and comfortable with nunit, but if xunit works for you then convert it.   Just don't have  both at once.

martincostello:
I'd definitely agree that we should only use _one_ test framework.

martincostello:
I couldn't get the tests to compile against .NET 4.5.1 (or .NET 4.5.2) for some reason, so they only target netcoreapp1.0 at the moment. If someone can work out what's missing we should be able to test both compilations.

JustEat.StatsD.Common.props
```diff
@@ -0,0 +1,21 @@
+﻿<Project>
```
slang25: I'll be copying this 😃

slang25:
Looks flawless 👍 
Would be interested to see what happens in Appveyor if you add back `net451`, works for me.

JustEat.StatsD.Common.props
```diff
@@ -0,0 +1,21 @@
...
+  <PropertyGroup>
+    <Authors>Just Eat</Authors>
+    <Company>Just Eat</Company>
+    <Copyright>Copyright (c) Just Eat 2015-2017</Copyright>
```
martincostello: Just realised could go one step further with this and use MSBuild to calculate the end year for the copyright...

martincostello:
@slang25 It went wonky on the command line and VS for me in different ways, so I'm not sure I want to wing it just yet :/

PR #18, opened 2017-03-10
Review dispose

AnthonySteele:
Review use of dispose, remove empty dispose pattern


martincostello:
Seems OK to me.

Possible accidentally flakey test from #17 might be making the CI in AppVeyor fail though (I just re-ran it and it was fine that time...).

AnthonySteele:
I can try to make that test more generous with times.

martincostello:
Would be good if you can while you're at it - it passed every single time I built it in the last PR so I didn't even realise it might need a bit of massaging.

PR #21, opened 2017-03-10
Rename the client / transport interface

AnthonySteele:
`IStatsDUdpClient` renamed to `IStatsDTransport`
Since we expect to have a TCP one


src/JustEat.StatsD/StatsDUdpTransport.cs
```diff
@@ -57,7 +57,7 @@ public bool Send(IEnumerable<string> metrics)
 
             try
             {
-                data.RemoteEndPoint = GetIPEndPoint();
+                data.RemoteEndPoint = GetIpEndPoint();
```
martincostello: Naming guidelines is usually to have two capitals if the acronym is only two letters in the framework generally, so how it was was correct.

src/JustEat.StatsD.Tests/Extensions/TimingConstants.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace JustEat.StatsD.Extensions
+{
+    public static class TimingConstants
+    {
+        public const int DelayMilliseconds = 400;
+
+        public static readonly TimeSpan Delta = TimeSpan.FromMilliseconds(DelayMilliseconds / 2);
```
martincostello: Nice.

martincostello:
50% of the way there on the timings...

PR #22, opened 2017-03-10
Rename the publisher

AnthonySteele:
It's the default publisher. it's the only publisher. calling it the "immediate" publisher is overly verbose.


martincostello:
Whilst unrelated to this specific change, I suggest we bump the version to 2.0.0 now before we forget later (or do it in a PR after this one).

PR #23, opened 2017-03-11
Constructor params object

AnthonySteele:
Replace all of the constructor params to `StatsDPublisher` - the `CultureInfo, string, int, string` with a new `StatsDConfiguration` type.

This is a breaking change for the config code of all clients.

The main reason that I prefer this style is that injecting a bunch of strings and ints into a constructor via IoC is error prone, tricky, verbose syntax that always varies from one IoC container to another.  So we replace [the primitive obsession](http://wiki.c2.com/?PrimitiveObsession) with strong typing, which IoC containers understand well. 

Off the back of that, the number of different constructor overloads and default params goes down since all the defaulting moves into this object. Should we need to add more config options, this will be simpler. 

I haven't looked at the `Host` configuration property which is stringly typed as either a host name or IP address, but that is a possibility afterwards.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,7 @@ An example of a very simple statsd publisher configuration, using the default va
 
 ```csharp
 
-var statsDConfig = new StatsDConfiguration
-{
-  HostNameOrAddress = "metrics_server.mycompany.com"
-};
+var statsDConfig = new StatsDConfiguration { Host = "metrics_server.mycompany.com" };
 var statsDPublisher = new StatsDPublisher(statsDConfig);
 ```
 
@@ -51,7 +48,7 @@ string statsdPrefix =  ConfigurationManager.AppSettings["statsd.prefix"];
 // it can be a preconfigured singleton instance
 var statsDConfig = new StatsDConfiguration
 {
-  HostNameOrAddress = statsdHostName,
+  Host = statsdHostName,
   Port = statsdPort,
   Prefix = statsdPrefix,
   Culture = CultureInfo.InvariantCulture
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ An example of a very simple statsd publisher configuration, using the default va
 ```csharp
 
 var statsDConfig = new StatsDConfiguration { Host = "metrics_server.mycompany.com" };
-var statsDPublisher = new StatsDPublisher(statsDConfig);
+IStatsDPublisher statsDPublisher = new StatsDPublisher(statsDConfig);
 ```
 
 An example of IoC in NInject for statsd publisher with values for all options, read from configuration:
--- a/README.md
+++ b/README.md
@@ -22,21 +22,44 @@ We use this within our components to publish [statsd](http://github.com/etsy/sta
 
 `IStatsDPublisher` is the interface that you will use in most circumstances. With this you can `Increment` or `Decrement` an event, and send values for a `Gauge` or `Timing`.
 
-The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. For the constructor parameters, you will need the statsd server host name. You can change the standard port (8125). You can also prepend a prefix to all stats. These values often come from configuration as the host name and/or prefix may vary between test and production environments.
+The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The constructor takes an instance of `StatsDConfiguration`. For the configuration's values, you will always need the statsd server host name or IP address. Optionally, you can change the standard port (8125). You can also prepend a prefix to all stats. These values often come from configuration as the host name and/or prefix may vary between test and production environments.
 
 #### Example of setting up a StatsDPublisher
 
-An example of Ioc in NInject for statsd publisher with values from configuration:
+An example of a very simple statsd publisher configuration, using the default values for most things.
+
 ```csharp
+
+var statsDConfig = new StatsDConfiguration
+{
+  HostNameOrAddress = "metrics_server.mycompany.com"
+};
+var statsDPublisher = new StatsDPublisher(statsDConfig);
+```
+
+An example of IoC in NInject for statsd publisher with values for all options, read from configuration:
+
+```csharp
+
+// read config values from app settings
 string statsdHostName =  ConfigurationManager.AppSettings["statsd.hostname"];
 int statsdPort = int.Parse(ConfigurationManager.AppSettings["statsd.port"]);
 string statsdPrefix =  ConfigurationManager.AppSettings["statsd.prefix"];
 
-Bind<IStatsDPublisher>().To<StatsDPublisher>()
-    .WithConstructorArgument("cultureInfo", CultureInfo.InvariantCulture)
-    .WithConstructorArgument("hostNameOrAddress",statsdHostName)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -22,21 +22,41 @@ We use this within our components to publish [statsd](http://github.com/etsy/sta
...
+The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The constructor takes an instance of `StatsDConfiguration`. For the configuration's values, you will always need the statsd server host name or IP address. Optionally, you can change the standard port (8125). You can also prepend a prefix to all stats. These values often come from configuration as the host name and/or prefix may vary between test and production environments.
 
 #### Example of setting up a StatsDPublisher
 
-An example of Ioc in NInject for statsd publisher with values from configuration:
+An example of a very simple statsd publisher configuration, using the default values for most things.
+
+```csharp
+
+var statsDConfig = new StatsDConfiguration { Host = "metrics_server.mycompany.com" };
```
martincostello: Maybe we could add some extension methods (later?) to make this prettier in the most simple case in a builder-ish pattern?

```csharp
new StatsDConfiguration().ForHost("metrics_server.mycompany.com");
```

README.md
```diff
@@ -22,21 +22,41 @@ We use this within our components to publish [statsd](http://github.com/etsy/sta
...
+IStatsDPublisher statsDPublisher = new StatsDPublisher(statsDConfig);
+```
+
+An example of IoC in NInject for statsd publisher with values for all options, read from configuration:
+
 ```csharp
+
+// read config values from app settings
```
martincostello: It would be nice to have an example for ASP.NET Core as well using the built-in IoC setup stuff and the strongly-typed options. I could update this later on if we want to add that as I've got experience of setting that stuff up already.

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Globalization;
+
+namespace JustEat.StatsD
+{
+    public class StatsDConfiguration
+    {
+        private const string SafeDefaultIsoCultureId = "en-US";
```
martincostello: I think the better name might be ```DefaultInvariantCulture```?

src/JustEat.StatsD/StatsDMessageFormatter.cs
```diff
@@ -10,22 +10,19 @@ namespace JustEat.StatsD
...
+        private static Random _random;
 
         private readonly CultureInfo _cultureInfo;
         private readonly string _prefix;
 
-        public StatsDMessageFormatter() : this(new CultureInfo(SafeDefaultIsoCultureId), prefix: "") {}
-
-        public StatsDMessageFormatter(string prefix = "") : this(new CultureInfo(SafeDefaultIsoCultureId), prefix) {}
-
         public StatsDMessageFormatter(CultureInfo ci, string prefix = "")
```
martincostello: As nice as optional parameters are in code you own, they cause a nightmare upstream as adding/removing them are breaking changes, which can be made by accident easily. I suggest we use constructor overload here and pass ```string.Empty``` in from a new overload.

martincostello: Lets call it ```culture``` instead of ```ci```.

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -8,18 +7,15 @@ namespace JustEat.StatsD
...
-        private static readonly CultureInfo SafeDefaultCulture = new CultureInfo(StatsDMessageFormatter.SafeDefaultIsoCultureId);
         private readonly StatsDMessageFormatter _formatter;
         private readonly IStatsDTransport _transport;
 
-        public StatsDPublisher(CultureInfo cultureInfo, string hostNameOrAddress, int port = 8125, string prefix = "")
+        public StatsDPublisher(StatsDConfiguration configuration)
         {
-            _formatter = new StatsDMessageFormatter(cultureInfo, prefix);
-            _transport = new StatsDUdpTransport(hostNameOrAddress, port);
+            _formatter = new StatsDMessageFormatter(configuration.Culture, configuration.Prefix);
```
martincostello: As this is library code, I think would be best to validate ```configuration``` is not ```null``` and ```throw new ArgumentNullException(nameof(configuration));```.

src/PerfTestHarness/Program.cs
```diff
@@ -12,7 +13,7 @@ private static void Main(string[] args)
         {
             var iterations = Enumerable.Range(1, 500000);
             var client = new StatsDUdpTransport(10, "localhost", 3128);
-            var formatter = new StatsDMessageFormatter();
+            var formatter = new StatsDMessageFormatter(CultureInfo.InvariantCulture);
```
martincostello: Actually, why don't we use this in the configuration object instead of ```en-US```?

martincostello:
@AnthonySteele Looks good in principal, but we might want to tweak it a bit as-per comments.

src/PerfTestHarness/Program.cs
AnthonySteele: IDK why it's like that. I'm happy to change to use `InvariantCulture` as the default.

AnthonySteele:
All good comments, I will work through them later.

README.md
AnthonySteele: Seems very possible.  Not really something that I need, but YMMV.

README.md
AnthonySteele: IMHo this tutorial code isn't really about "How do I read values from config?" so maybe we could skip that part entirely? Or is it necessary?

src/PerfTestHarness/Program.cs
AnthonySteele: `InvariantCulture` is now the default

src/JustEat.StatsD/StatsDMessageFormatter.cs
AnthonySteele: We can add all the constructor overloads back in, though IMHO that's not necessary as this class is mostly used in the internals of the lib, and in the tests, so client code calling the other constructors will be extremely rare.

README.md
martincostello: I'll have a ponder about what we might want to change it to, if at all, leave it as is for now I reckon then.

martincostello:
LGTM.

README.md
AnthonySteele: I think we can address this later. If we do want to talk about config, there's likely more value in talking about the new .Net Core config and IoC bits as they are much less well known.

AnthonySteele:
If we are looking at validating inputs, then the `configuration.Host` string should not be empty - or is that ever done to turn off statting in test environments?

martincostello:
I'm not sure. I'd have thought the best way to do that kind of behaviour would be to self-test your configuration and not register it in IoC, rather than rely on the implementation to treat a special value as meaning "off". In that case, I guess we should validate that a host has been specified too.

AnthonySteele:
I think that this PR is ready.
There are other things that come off the back of it, but this is big enough.

PR #25, opened 2017-03-15
Ip end point source

AnthonySteele:
A simpler approach to resolving Ip+port endpoints
Moves some code out of `StatsDUdpTransport` so that it can be reused in other transports
And does not keep the DNS mapper around when it's not needed.

The outstanding question is - how was DNS caching working before? where did the `endpointCacheDuration` value come from, was the client supposed to set this up themselves if they wanted it? I don't think that this was simple before, but can it be simple now? What would be a reasonable value for a default DNS cache time?


src/JustEat.StatsD/EndpointLookups/CachedIpEndpointSource.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// </summary>
+    public class CachedIpEndpointSource : IPEndPointSource
+    {
+        private IPEndPoint _cachedValue;
+        private DateTime _expiry;
+        private readonly int _cacheDurationSeconds;
+
+        private readonly IPEndPointSource _inner;
+
+        public CachedIpEndpointSource(IPEndPointSource inner, int cacheDurationSeconds)
```
martincostello: Personally I'd prefer a TimeSpan so calling code looks a bit more readable:

```csharp
new CachedIpEndpointSource(inner, 5 * 60);
```

vs.

```csharp
new CachedIpEndpointSource(inner, TimeSpan.FromMinutes(5));
```

It also allows the consumer arbitrary precision (though why you'd cache for less than one second I don't know).

src/JustEat.StatsD/EndpointLookups/DnsLookupIpEndpointSource.cs
```diff
@@ -0,0 +1,28 @@
...
+        private readonly string _hostName;
+        private readonly int _port;
+
+        public DnsLookupIpEndpointSource(string hostName, int port)
+        {
+            _hostName = hostName;
+            _port = port;
+        }
+
+        public IPEndPoint Endpoint => new IPEndPoint(GetIpAddressOfHost(_hostName), _port);
```
martincostello: This feels more like a method than a property to me as it does stuff that's non-trivial, plus if you accessed the value via the property more than once in the same method you'd get different values, which might be unexpected.

src/JustEat.StatsD/EndpointLookups/DnsLookupIpEndpointSource.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        private IPAddress GetIpAddressOfHost(string hostName)
+        {
+            var endpoints = Dns.GetHostAddressesAsync(hostName)
+                .GetAwaiter().GetResult();
+            return endpoints[0];
```
martincostello: Verify there's at least one so we don't get ```ArgumentOutOfRangeException```?

src/JustEat.StatsD/EndpointLookups/EndpointParser.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Net;
+
+namespace JustEat.StatsD.EndpointLookups
+{
+    public static class EndpointParser
+    {
+        public static IPEndPointSource MakeEndPointSource(string host, int port, int? endpointCacheDuration)
```
martincostello: Use ```TimeSpan?``` as per other comment?

src/JustEat.StatsD/EndpointLookups/EndpointParser.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            if (string.IsNullOrWhiteSpace(host))
+            {
+                throw new ArgumentException(nameof(host));
```
martincostello: Specify a message.

src/JustEat.StatsD/EndpointLookups/IPEndPointSource.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Net;
+
+namespace JustEat.StatsD.EndpointLookups
+{
+    public interface IPEndPointSource
+    {
+        IPEndPoint Endpoint { get; }
```
martincostello: I think this should be a method as mentioned above.

src/JustEat.StatsD/StatsDUdpTransport.cs
```diff
@@ -15,29 +14,16 @@ public class StatsDUdpTransport : IStatsDTransport
...
-        public StatsDUdpTransport(int endpointCacheDuration, string hostNameOrAddress, int port)
-            : this(new CachedDnsEndpointMapper(new DnsEndpointProvider(), endpointCacheDuration), hostNameOrAddress, port) {}
-
-        private StatsDUdpTransport(IDnsEndpointMapper endpointMapper, string hostNameOrAddress, int port)
+        public StatsDUdpTransport(IPEndPointSource endPointSource)
         {
-            _endPointMapper = endpointMapper;
-            _hostNameOrAddress = hostNameOrAddress;
-            _port = port;
+            _endpointSource = endPointSource;
```
martincostello: Null check.

src/JustEat.StatsD/EndpointLookups/DnsLookupIpEndpointSource.cs
AnthonySteele: Fair enough, done.

src/JustEat.StatsD/EndpointLookups/DnsLookupIpEndpointSource.cs
AnthonySteele: 👍

src/JustEat.StatsD/StatsDUdpTransport.cs
AnthonySteele: 👍

src/JustEat.StatsD/EndpointLookups/CachedIpEndpointSource.cs
AnthonySteele: 👍

src/JustEat.StatsD/EndpointLookups/SimpleIpEndpoint.cs
```diff
@@ -7,11 +7,16 @@ namespace JustEat.StatsD.EndpointLookups
...
+        private readonly IPEndPoint _value;
+
         public SimpleIpEndpoint(IPEndPoint value)
         {
-            Endpoint = value;
+            _value = value;
         }
 
-        public IPEndPoint Endpoint { get; }
+        public IPEndPoint GetEndpoint()
```
martincostello: Minor stylistic point: maybe use ```public IPEndPoint GetEndpoint() => _value;``` ?

PR #26, opened 2017-03-18
Surface the Transport

AnthonySteele:
No breaking change in the simple case, but this gives options for injecting a custom transport into the `StatsDPublisher`.

My thinking is that `StatsDPublisher` is the top of the object hierarchy, so that's where assembling it all should happen,  and the `StatsDUdpTransport` can be dumber - it doesn't need to know about how to make an endpoint source, it just gets given one. So when there's a second transport, this endpoint-making code doesn't get duplicated.

`StatsDPublisher` surfaces a constructor with a `StatsDConfiguration` and a `IStatsDTransport` for the case where you use a different transport.

An alternative to this would be an enum in the config object, e.g. :
```csharp
// a different way to do it
var statsDConfig = new StatsDConfiguration 
  { 
     Host = "metrics_server.mycompany.com",
     Transport = StatsDTransport.Ip
};
IStatsDPublisher statsDPublisher = new StatsDPublisher(statsDConfig);
```

And then the `StatsDPublisher` constructor would have to create the appropriate  `IStatsDTransport` for the enum value.

### DIFF
--- a/src/JustEat.StatsD/StatsDPublisher.cs
+++ b/src/JustEat.StatsD/StatsDPublisher.cs
@@ -41,7 +41,8 @@ namespace JustEat.StatsD
 
             _formatter = new StatsDMessageFormatter(configuration.Culture, configuration.Prefix);
 
-            var endpointSource = EndpointParser.MakeEndPointSource(configuration.Host, configuration.Port, StatsDConfiguration.DefaultDnsLookupInterval);
+            var endpointSource = EndpointParser.MakeEndPointSource(
+                configuration.Host, configuration.Port, configuration.DnsLookupInterval);
             _transport = new StatsDUdpTransport(endpointSource);
         }
 
--- a/src/JustEat.StatsD/StatsDPublisher.cs
+++ b/src/JustEat.StatsD/StatsDPublisher.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using JustEat.StatsD.EndpointLookups;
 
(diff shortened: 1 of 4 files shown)


AnthonySteele:
This PR is read for review / merge.

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -23,7 +40,9 @@ public StatsDPublisher(StatsDConfiguration configuration)
             }
 
             _formatter = new StatsDMessageFormatter(configuration.Culture, configuration.Prefix);
-            _transport = new StatsDUdpTransport(configuration.Host, configuration.Port, null);
+
+            var endpointSource = EndpointParser.MakeEndPointSource(configuration.Host, configuration.Port, StatsDConfiguration.DefaultDnsLookupInterval);
```
martincostello: Seems a little odd to me here that we have an instance of ```StatsDConfiguration``` here, but we don't use it to get the endpoint lookup timeout and instead reference a static property. I get that it's not always relevant, it just _seems_ odd.

AnthonySteele: Yeah, we could put an instance property on StatsDConfiguration with that as the default value.

AnthonySteele:
^ Done.  

The fields of `StatsDConfiguration` all need doc comments for intellisense when using it, but that's a different PR.

PR #27, opened 2017-03-20
docs for tooltips

AnthonySteele:
Docs for tooltips on the StatsDConfiguration.
These should come up in intellisense when using this object.

### DIFF
--- a/src/JustEat.StatsD/StatsDConfiguration.cs
+++ b/src/JustEat.StatsD/StatsDConfiguration.cs
@@ -8,13 +8,35 @@ namespace JustEat.StatsD
         public const int DefaultPort = 8125;
         public static readonly TimeSpan DefaultDnsLookupInterval = TimeSpan.FromMinutes(5);
 
+        /// <summary>
+        /// The host name or IP address of the statsD server. 
+        /// This field must be set. 
+        /// </summary>
         public string Host { get; set; }
 
+        /// <summary>
+        /// The port on the satsD server that stas are sent to.
+        /// Default is the standard port (8125).
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -8,13 +8,35 @@ public class StatsDConfiguration
...
         public static readonly TimeSpan DefaultDnsLookupInterval = TimeSpan.FromMinutes(5);
 
+        /// <summary>
+        /// The host name or IP address of the statsD server. 
+        /// This field must be set. 
+        /// </summary>
         public string Host { get; set; }
 
+        /// <summary>
+        /// The port on the satsD server that stas are sent to.
```
martincostello: Typo: "stas"

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -8,13 +8,35 @@ public class StatsDConfiguration
...
+        /// Default is the standard port (8125).
+        /// </summary>
         public int Port { get; set; } = DefaultPort;
 
+        /// <summary>
+        /// Length of time to cache the host name to Ip address lookup.
```
martincostello: Use "IP".

slang25:
I'm a little late to the party, but I think your last PR isn't as thread-safe as it previously was (or should be). I'll take another look when I'm more awake =)

AnthonySteele:
@slang25 It's likely, I haven't given a lot of thought to that. Are you talking about concurrency in the `CachedIpEndpointSource`, or something else?

AnthonySteele:
Merge this one? It should be good.

PR #28, opened 2017-04-02
Review transport

AnthonySteele:
Review the UDP transport
Remove the `System.Diagnostics.CodeAnalysis` thing, we don't use it.
Rename - Remove redundant "Statsd" in the name `JustEat.StatsD.StatsDUdpTransport`.  What else could it transport?


martincostello:
Seems legit.

PR #30, opened 2017-04-07
Add an IP version of the transport

AnthonySteele:
Useful for lambdas

Want review before merging, on suitability for using in C# AWS lambdas.

### DIFF
--- a/src/JustEat.StatsD/IpTransport.cs
+++ b/src/JustEat.StatsD/IpTransport.cs
@@ -1,0 +1,59 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Net.Sockets;
+using JustEat.StatsD.EndpointLookups;
+
+namespace JustEat.StatsD
+{
+    public class IpTransport : IStatsDTransport
+    {
+        private readonly IPEndPointSource _endpointSource;
+
+        public IpTransport(IPEndPointSource endPointSource)
+        {
+            if (endPointSource == null)
+            {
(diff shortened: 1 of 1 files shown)


src/JustEat.StatsD/IpTransport.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private readonly IPEndPointSource _endpointSource;
+
+        public IpTransport(IPEndPointSource endPointSource)
+        {
+            if (endPointSource == null)
+            {
+                throw new ArgumentNullException(nameof(endPointSource));
+            }
+            _endpointSource = endPointSource;
```
martincostello: Now we're on VS 2017 for this, could write this as:

```csharp
```

AnthonySteele: We _can_ do that in a couple of places. it seemed a little cutting-edge for me at the time, but if you like I can change it.

martincostello: Just a suggestion as this is new. Not essential, but would make it a bit more terse.

AnthonySteele: Maybe in  a seperate PR, as I would apply it across all classes.

AnthonySteele: I suggest that a review of using new syntax throughout be left to a later PR. But let's actually do it some time soon.

martincostello: 👍

PR #31, opened 2017-04-10
Review use of new syntax

AnthonySteele:
use some new syntax - `throw` expressions, `nameof` operator and an `=>` property get
refactoring guided by resharper
Should be the same only less code

### DIFF
--- a/src/JustEat.StatsD/DisposableTimer.cs
+++ b/src/JustEat.StatsD/DisposableTimer.cs
@@ -14,11 +14,11 @@ namespace JustEat.StatsD
 
         public DisposableTimer(IStatsDPublisher publisher, string statName)
         {
-            _publisher = publisher ?? throw new ArgumentNullException("publisher");
+            _publisher = publisher ?? throw new ArgumentNullException(nameof(publisher));
 
             if (string.IsNullOrEmpty(statName))
             {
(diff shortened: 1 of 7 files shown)


src/JustEat.StatsD/DisposableTimer.cs
```diff
@@ -14,17 +14,13 @@ internal class DisposableTimer : IDisposableTimer
 
         public DisposableTimer(IStatsDPublisher publisher, string statName)
         {
-            if (publisher == null)
-            {
-                throw new ArgumentNullException("publisher");
-            }
+            _publisher = publisher ?? throw new ArgumentNullException("publisher");
```
martincostello: Use ```nameof(publisher)```.

src/JustEat.StatsD/DisposableTimer.cs
```diff
@@ -14,17 +14,13 @@ internal class DisposableTimer : IDisposableTimer
...
 
             if (string.IsNullOrEmpty(statName))
             {
                 throw new ArgumentNullException("statName");
```
martincostello: Might as well change this one too.

martincostello:
LGTM.

slang25:
Very nice

PR #32, opened 2017-04-12
Remove junk serialisation attr

AnthonySteele:
Remove a junk attribute.

- the `StatsDMessageFormatter` never needs to be serialized anyway, it is behaviour not data.

- because the lib is compiled for `netstandard1.6` and `net451`, it looks to me like the `#if NET45`  is actually never true. You can put any text in that block and it still builds fine, twice over.  `#if` is pretty tricksy.


martincostello:
LGTM.

PR #37, opened 2017-08-10
Remove Trace.Information() calls

martincostello:
There have been complaints that StatsD is too noisy to the trace stream in debuggers.

In the short term, address this by only doing these calls in DEBUG. In the long term, we should log to an abstraction so consumers can control this.

`Trace.Debug()` would have been nicer, but alas, it does not exist.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,28 @@
+{
+   // Use IntelliSense to find out which attributes exist for C# debugging
+   // Use hover for the description of the existing attributes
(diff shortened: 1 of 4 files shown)


AnthonySteele:
I have no objections. Why not remove them altogether?

martincostello:
I can remove them entirely if that suits - putting them around an `#if DEBUG` was just a conservative middle-ground.

AnthonySteele:
:shipit:

PR #38, opened 2017-10-13
Use an OnError function to report exceptions

AnthonySteele:
This is what it looks like if you remove the use of the `System.Diagnostics.Trace` things, which have caused issues for users.
The questions are though
- what to do with exceptions
- was this a good thing to do with them in the first place?

### DIFF
--- a/src/JustEat.StatsD/StatsDConfiguration.cs
+++ b/src/JustEat.StatsD/StatsDConfiguration.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Globalization;
 
 namespace JustEat.StatsD
@@ -38,5 +38,12 @@ namespace JustEat.StatsD
         /// Default is InvariantCulture.
         /// </summary>
         public CultureInfo Culture { get; set; } = CultureInfo.InvariantCulture;
+
+        /// <summary>
+        /// Function to receive notificaion of any exceptions
+        /// Returns: True if it has handled the exception and no further action is needed, 
+        /// false if it should be re-thrown
+        /// </summary>
+        public Func<Exception, bool> OnError { get; set; } = e => true;
     }
 }
--- a/src/JustEat.StatsD/StatsDConfiguration.cs
+++ b/src/JustEat.StatsD/StatsDConfiguration.cs
@@ -40,10 +40,12 @@ namespace JustEat.StatsD
         public CultureInfo Culture { get; set; } = CultureInfo.InvariantCulture;
 
         /// <summary>
-        /// Function to receive notificaion of any exceptions
-        /// Returns: True if it has handled the exception and no further action is needed, 
-        /// false if it should be re-thrown
+        /// Function to receive notification of any exceptions
+        /// This function should return:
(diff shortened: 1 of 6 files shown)


martincostello:
Maybe as a separate PR we should add a callback/interface or something that allows a consumer to optionally do something with an exception that is thrown when trying to send metrics?

AnthonySteele:
@martincostello I guess, maybe an Action<Exception> on the settings object?

martincostello:
@AnthonySteele Or possibly `Func<Exception, bool>`? Then we could optionally throw based on the result returned.

AnthonySteele:
This is a bigger PR now, what do you think? Separate it out maybe?

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -38,5 +38,12 @@ public class StatsDConfiguration
         /// Default is InvariantCulture.
         /// </summary>
         public CultureInfo Culture { get; set; } = CultureInfo.InvariantCulture;
+
+        /// <summary>
+        /// Function to receive notificaion of any exceptions
```
martincostello: Typo: _notificaion_ should be _notification_.

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -38,5 +38,12 @@ public class StatsDConfiguration
...
+        /// Returns: True if it has handled the exception and no further action is needed, 
+        /// false if it should be re-thrown
```
martincostello: "be thrown" rather than "re-thrown" as we're using `throw`?

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -38,5 +38,12 @@ public class StatsDConfiguration
...
+        /// </summary>
+        public Func<Exception, bool> OnError { get; set; } = e => true;
```
martincostello: Document the default behaviour in `<remarks></remarks>`?

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -40,71 +42,88 @@ public StatsDPublisher(StatsDConfiguration configuration)
...
+
+        private void Send(string metric)
+        {
+            try
+            {
+                _transport.Send(metric);
+            }
+            catch (Exception ex)
+            {
+                var handled = _onError(ex);
```
martincostello: There's no guarantee that `_onError` is non-`null` here. Maybe use a construct like this instead?

```csharp
bool handled = _onError?.Invoke(ex) ?? true;
```

AnthonySteele: If we use that code, then the default in `StatsDConfiguration` might as well be null. Seems simpler.

martincostello:
Is it possible to add a test for the `onError` stuff, or would we need to refactor it too much to abstract it out to force the exceptions to be thrown?

martincostello:
Nice.

Issue #42, opened 2018-02-20
SendBufferSize of 0 causes SocketException in OS-X


simonjduff:
Repro:

```
using System;
using System.Net.Sockets;
namespace simon
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
		var socket = new UdpClient{Client = {SendBufferSize=0}};
        }
    }
}
````

when run results in

````
Unhandled Exception: System.Net.Sockets.SocketException: Invalid argument
   at System.Net.Sockets.Socket.UpdateStatusAfterSocketErrorAndThrowException(SocketError error, String callerName)
   at System.Net.Sockets.Socket.SetSocketOption(SocketOptionLevel optionLevel, SocketOptionName optionName, Int32 optionValue, Boolean silent)
   at System.Net.Sockets.Socket.SetSocketOption(SocketOptionLevel optionLevel, SocketOptionName optionName, Int32 optionValue)
   at System.Net.Sockets.Socket.set_SendBufferSize(Int32 value)
   at simon.Program.Main(String[] args) in /Users/tom.maton/Dev/simon/Program.cs:line 10
````

https://github.com/justeat/JustEat.StatsD/blob/df3aff460c4bb71f8a14042ea97c2df1b8f51fae/src/JustEat.StatsD/UdpTransport.cs#L51

martincostello:
If we can't write a test that repros this on Linux in Travis CI, we could ask them to enable macOS builds in parallel as well as part of the Travis CI and test this on three platforms instead of two.

simonjduff:
I've just tried the same in an Ubuntu 16.04 VM with no repro, so it looks like just an OS-X issue.
Obviously unlikely to affect production applications, but this issue interrupts the developer experience when working with dotnet core on OS-X.

martincostello:
Cool - I'll see if I can get Travis to enable macOS builds.

martincostello:
Request sent to `support@travis-ci.com`.

PR #43, opened 2018-02-20
Fix UdpTransport on macOS

martincostello:
To start with, this PR just adds a test that sending a metric does not throw an exception. Once Travis CI enable macOS parallel builds for us, this test should fail.

After that, we can apply the suggested fix and verify.

Relates to #42.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 sudo: required
 dist: trusty
 
+os:
+  - linux
+  - osx
+
 env:
   global:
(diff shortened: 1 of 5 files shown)


martincostello:
🥁 
image

martincostello:
Failure "successful" on macOS: https://travis-ci.org/justeat/JustEat.StatsD/jobs/343973341#L327-L351

martincostello:
Fixed - I think this is ready to go.

PR #44, opened 2018-02-22
Detail all settings in the readme

AnthonySteele:


martincostello:
Maybe use backticks around the default values and their type names, but otherwise looks good 👍

PR #46, opened 2018-02-22
Fix package description

### LINKED ISSUE #45: Package description for NuGet is wrong
Oops. Looks like we broke this since 1.1.0.

image

martincostello:
* Add a package description/summary for NuGet.
  * Bump the version to `2.0.1`.

Resolves #45.

### DIFF
--- a/JustEat.StatsD.Common.props
+++ b/JustEat.StatsD.Common.props
@@ -1,4 +1,4 @@
-﻿<Project>
+<Project>
   <PropertyGroup>
     <Authors>Just Eat</Authors>
     <Company>Just Eat</Company>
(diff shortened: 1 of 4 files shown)


martincostello:
@slang25 What black magic sets the version number? I can't find it and the version is still coming out as 2.0.0 in the CI builds.

stuart-lang:
Looks like this: https://github.com/justeat/JustEat.StatsD/blob/master/JustEat.StatsD.Common.props#L18

stuart-lang:
Here's what we do in JustSaying: https://github.com/justeat/JustSaying/blob/master/version.props

martincostello:
🤦‍♂️ dammit - I looked for that and completely missed it...

martincostello:
We can make the approach consistent between the repos in another PR.

PR #49, opened 2018-02-24
Service registration extensions

martincostello:
Some initial ideas on baking service registration extensions for .NET Core into the library itself to avoid the need for repetitive boilerplate in our consuming applications.

I've included several different flavours of registration method, maybe _too_ many, but we can always remove some if we think it's a bit too excessive.

Main use cases I've tried to cater for are:
  * Simple registering of the default required types, but with the step to configure the host left to the integrator to achieve by other means, such as using `Configure()` from the `Microsoft.Extensions.Configuration` package.
  * Simple registration with a string for the host name and an optional prefix.
  * Registration from a provided `StatsDConfiguration` instance.
  * Registration using a callback delegate to configure a `StatsDConfiguration` object passed to the caller, similar to `Configure()`.
  * Registration with a factory method to provide a `StatsDConfiguration` for us to use provided by the callee.
  * Variations on the above but with `IServiceProvider` as a delegate parameter so the integrator can lazily-resolve their other services  to populate/provide a `StatsDConfiguration` instance.

Use the included tests as pseudo-examples of use cases. Proper documentation can be added later after review/refinement.

For now, I've intentionally only supported this with `netstandard1.6` to avoid polluting `net451` consumers with all the extra libraries from ".NET Standard Land" full framework projects would otherwise pull in and cause a mess/bloat. We can revisit that if there's a nice way to make it available for them without introducing a tonne of additional dependencies.

### DIFF
--- a/JustEat.StatsD.Common.props
+++ b/JustEat.StatsD.Common.props
@@ -15,7 +15,7 @@
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>$(PackageProjectUrl).git</RepositoryUrl>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <VersionPrefix>2.0.1</VersionPrefix>
+    <VersionPrefix>2.1.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
   </PropertyGroup>
 </Project>
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 os: Visual Studio 2017
-version: 2.0.1.{build}
+version: 2.1.0.{build}
 configuration: Release
 environment:
   CLI_VERSION: 2.1.4
--- a/src/JustEat.StatsD.Tests/JustEat.StatsD.Tests.csproj
+++ b/src/JustEat.StatsD.Tests/JustEat.StatsD.Tests.csproj
@@ -11,6 +11,7 @@
     <Content Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="1.0.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.6.0" />
     <PackageReference Include="Moq" Version="4.8.1" />
     <PackageReference Include="Shouldly" Version="3.0.0" />
--- a/src/JustEat.StatsD.Tests/JustEat.StatsD.Tests.csproj
+++ b/src/JustEat.StatsD.Tests/JustEat.StatsD.Tests.csproj
@@ -11,7 +11,7 @@
     <Content Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="1.0.0" />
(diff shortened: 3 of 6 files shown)


AnthonySteele:
Good idea, I suggest that we start with a minimal set of extensions for only NetStandard/NetCore apps, and which cover the actual scenarios where we are using StatsD v2 in our apps - there are a couple of these already.

martincostello:
I might also update this to register the endpoint source as well, as that’d make registering a custom `IStatsDTransport` easier (I think).

martincostello:
Plugging this into an internal application, I got the StatsD registration down to:

```csharp
services.AddSingleton<IStatsDTransport, CustomUdpTransport>();
services.AddStatsD(
    (provider) =>
    {
        var options = provider.GetRequiredService<CustomOptions>().StatsD;
        var logger = provider.GetRequiredService<ILogger<StatsDConfiguration>>();

        return new StatsDConfiguration()
        {
            Host = options.Hostname,
            Port = options.Port,
            Prefix = options.Prefix,
            OnError = (error) =>
            {
                logger.LogWarning(error, "Failed to publish metric to StatsD. {Message}", error.Message);
                return true;
            },
        };
    });
```

AnthonySteele:
Is it time to look at feeding in the `IStatsDTransport` via the config object?

martincostello:
I don't think so, I don't see that as configuration, personally. It's more DI registration.

stuart-lang:
This is really nice @martincostello. With the 2 examples you gave I think I slightly prefer the `Action<StatsDConfiguration>` but not strongly enough to make a case for it.

Would it be worth making this a separate package, I know that it is an abstractions package that is only present in the netstandard target, but it still makes me feel slightly uncomfortable having it all in one package.

martincostello:
It's quite common for netstandard stuff to ship with its own `IServiceCollection` extensions nowadays though. It would seem a bit weird to me in a netcoreapp application to have to install a non-obvious extra package just get the easiest "plug-in-and-go" experience.

stuart-lang:
Oh right, I thought the opposite, but I'm thinking of MediatR and AutoMapper. Do you have examples?

martincostello:
To be honest I'm thinking of more the core framework packages themselves (`HttpClientFactory` is doing this), rather than 3rd party things, but a whole extra package for 1 class seems like overkill.

stuart-lang:
Yeah, that's a fair point 👍

martincostello:
So I think the two outstanding questions I have are:
  1. Which overloads do we want to drop?
  1. Does this justify a bump for the next version of `2.1.0` instead of `2.0.1`?

AnthonySteele:
it should be version `2.1.0`
> MINOR version when you add functionality in a backwards-compatible manner
https://semver.org/

martincostello:
Version bumped, so which overloads do we want if not all?

My suggestions are:
  1. `AddStatsD(string host, string prefix = null)` - Simplest use case to get things usable.
  1. `AddStatsD(Func<IServiceProvider, StatsDConfiguration> configurationFactory)` - Allows you to provide your own configuration object, with the service container available to allow you to resolve any other dependencies you might need to be able to do that. This is the one in the first registration example above.
  1. `AddStatsD()` - One-liner that configures everything you need but leaves you to populate the configuration yourself somehow, either by registering it yourself beforehand, or using `Configure()`.

martincostello:
@AnthonySteele @stuart-lang *poke*

martincostello:
@AnthonySteele Have trimmed the `AddStatsD(...)` overloads as discussed this morning.

PR #50, opened 2018-03-06
Update README for service registration

martincostello:
* Update the `README` with examples of using service registration from #49.
  * Add link to `CHANGELOG` for once the release is published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,6 @@ services.AddStatsD(
     (provider) =>
     {
         var options = provider.GetRequiredService<MyOptions>().StatsD;
-        var logger = provider.GetRequiredService<ILogger<StatsDConfiguration>>();
 
         return new StatsDConfiguration()
         {
--- a/README.md
+++ b/README.md
@@ -30,13 +30,56 @@ The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The 
 
 An example of a very simple StatsD publisher configuration, using the default values for most things.
 
-```csharp
+#### .NET Core
 
+```csharp
+// Service registration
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,13 +30,56 @@ The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The
...
+#### .NET Core
+
+An example of registering StatsD dependencies using `IServiceCollection`:
+
+```csharp
+services.AddStatsD(
+    (provider) =>
+    {
+        var options = provider.GetRequiredService<MyOptions>().StatsD;
+        var logger = provider.GetRequiredService<ILogger<StatsDConfiguration>>();
```
AnthonySteele: Logger is not used in this example. Perahs should be in the `LogError(ex)` line?

martincostello: Good spot. It used to be, but then I took the code from the existing example and simplified it.

README.md
```diff
@@ -30,13 +30,55 @@ The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The
...
+{
+    public MyService(IStatsDPublisher statsPublisher)
+    {
+        // Use the publisher
+    }
+}
+
+```
+
+#### .NET 4.5.1
```
AnthonySteele: You can still do this .Net core. You just aren't taking advantage of the service registration bits.

martincostello: Suggested alternate heading?

AnthonySteele: Hmmm, maybe

#### Simplest example
No service registration or Ioc. Works in .NET 4.5.1 and .NET core.

PR #53, opened 2018-04-13
Adding Gauge methods that support double as a parameter

tererivas:
Adding Gauge methods that support double as a parameter

Added Gauge methods in Publisher and Formatter classes that take a double as a parameter. I need to send fractional values to StatsD and a long value didn't allow me to do that.

### DIFF
--- a/src/JustEat.StatsD.Tests/Extensions/FakeStatsPublisher.cs
+++ b/src/JustEat.StatsD.Tests/Extensions/FakeStatsPublisher.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 
 namespace JustEat.StatsD.Extensions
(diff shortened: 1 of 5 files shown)


AnthonySteele:
Looks fine in principle, but the `FakeStatsPublisher` also needs the new gauge methods for a green build.

martincostello:
We’ll also need to bump the version to `3.0.0` as this is a breaking interface change.

martincostello:
Also, please update the relevant unit test class to exercise the new members.

tererivas:
Ok, I will do the requested changes and update the PR. Thank you!

tererivas:
Question, you have 3 unit tests and all of them use the gauge with a long parameter, should I create 1 more unit tests that uses a double parameter instead, similar to GaugeMetricsAreFormattedCorrectly() or should I create 3 new unit tests, just like the ones you currently have but using a double parameter instead?

AnthonySteele:
There are 2 new method overloads, so I would expect around 3 new tests in order to cover both of them, and the cultureInfo part.

tererivas:
@martincostello mentioned above that the version should probably be bumped up to version to 3.0.0. Is that something I should do? Also, when will these changes be merged into master? Thank you.

AnthonySteele:
No, there is no need to change the version number in this PR. Martin is out of the office today, but we will see about getting it merged soon.

stuart-lang:
@tererivas Congrats on your first-time PR 👏

tererivas:
Thank you!!

PR #55, opened 2018-04-20
Simplify send

AnthonySteele:

remove the `Send(IEnumerable<string> metrics)` overload
- we don't use it, even internally, and there's lots of code there to deal with it which can be deleted.
- And makes the single-metric case that we do use slightly better as it doesn't have to be wrapped in an array of one element, and then iterated.

**done** Review with @slang25 , we wanted to benchmark & optimise and decided to delete instead
Will later want to optimise remaining code using `Span<T>` 

**done**  Needs testing on QA servers to verify that Ip and UDP transports still work as expected.


src/JustEat.StatsD/IpTransport.cs
```diff
@@ -15,21 +15,14 @@ public IpTransport(IPEndPointSource endPointSource)
...
-            var packets = metrics.ToMaximumBytePackets();
+            var bytes = Encoding.UTF8.GetBytes(metric);
 
             using (var socket = MakeIpDatagramSocket())
             {
-                foreach (var packet in packets)
-                {
-                    socket.SendTo(packet, endpoint);
-                }
+                socket.Connect(endpoint);
```
martincostello: Is this call to `Connect()` required given it wasn't there before?

src/JustEat.StatsD/IpTransport.cs
```diff
@@ -15,21 +15,14 @@ public IpTransport(IPEndPointSource endPointSource)
...
         {
             var endpoint = _endpointSource.GetEndpoint();
-            var packets = metrics.ToMaximumBytePackets();
+            var bytes = Encoding.UTF8.GetBytes(metric);
 
             using (var socket = MakeIpDatagramSocket())
             {
-                foreach (var packet in packets)
-                {
-                    socket.SendTo(packet, endpoint);
```
martincostello: Even if we only sent one metric, this code would have split the string into 512-byte packets if required. Is that something we don't need to worry about anymore?

src/JustEat.StatsD/UdpTransport.cs
```diff
@@ -23,45 +19,31 @@ public UdpTransport(IPEndPointSource endPointSource)
...
             {
-                udpClient.Client.Connect(data.RemoteEndPoint);
-                udpClient.Client.SendPacketsAsync(data);
+                socket.Connect(endpoint);
+                socket.Send(bytes);
             }
         }
 
-        public UdpClient GetUdpClient()
+        public Socket GetUdpClient()
```
martincostello: `CreateSocket()`?

martincostello: Might be worth cross-referencing this changes with the implementation I have internally that pools `UdpClient` (or is it `Socket`) that I created when profiling seemed to suggest that opening and closing lots of sockets created a fair bit of CPU overhead. I never got conclusive enough results in isolation to suggest it was a change worth making in general for the library.

src/JustEat.StatsD/IpTransport.cs
slang25: Rather than use `SendTo`, you can use `Connect` & `Send` separately, this means we can take advantage of the overloads on `Send` that support `Span<T>` and family

src/JustEat.StatsD/IpTransport.cs
AnthonySteele: As I understand it, you can't split one metric over multiple packets. The packets are independent of each other and there's no state shared between them. Each packet should contain one or more valid metrics.

In `PacketBuilder` line 43-ish, I think it means that if the metric text is larger than the desired size for a packet, then you get an extra-large packet and there's no way around that in this code.

src/JustEat.StatsD/IpTransport.cs
AnthonySteele: We have changed `SendTo` to `Connect; Send` which we think is equivalent. Can change back for now if it's not.

src/JustEat.StatsD/IpTransport.cs
martincostello: OK cool. It just looked like that was how it worked at a high-level.

martincostello:
Nice work with the allocations.

AnthonySteele:
We think that the huge allocation in the "before, UDP" case is likely due to all the `SimpleObjectPool` code, which is a one-off upfront cost of pre-allocating, so we haven't saved as much as it seems like.

martincostello:
Maybe manually set the BenchmarkDotNet iteration count to be something really high so it balances out (or not)? I would've though the old code would definitely have more allocations due to the use of `yield` etc to create the packets.

src/JustEat.StatsD/UdpTransport.cs
```diff
@@ -29,7 +29,7 @@ public void Send(string metric)
             }
         }
 
-        public Socket CreateSocket()
+        private static Socket CreateSocket()
```
martincostello: 👍 - I saw this during 2.x work, but didn't change it as it would've been a breaking change (and then forgot about it).

AnthonySteele:
So I have changed the benchamrk.net test to send a "warmup" metric in the `GlobalSetup` and re-run.

On master:

 Method |     Mean |    Error |   StdDev |   Gen 0 | Allocated |
------- |---------:|---------:|---------:|--------:|----------:|
 RunUdp | 299.6 us | 3.906 us | 3.653 us | 44.9219 | 132.35 KB |
  RunIp | 224.1 us | 2.970 us | 2.632 us |  0.9766 |   3.62 KB |
  

On this branch:


 Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|-------:|----------:|
 RunUdp | 269.7 us | 1.345 us | 1.123 us | 0.4883 |   1.85 KB |
  RunIp | 252.7 us | 2.016 us | 1.886 us | 0.4883 |   1.84 KB |
  
RunIp is slightly slower, so with the additional change the that we go back t using `SendTo` instead of `Connect; Send` for both:


 Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
------- |---------:|---------:|---------:|-------:|----------:|
 RunUdp | 235.7 us | 3.867 us | 3.617 us | 0.4883 |   1.85 KB |
  RunIp | 218.0 us | 3.573 us | 3.343 us | 0.4883 |   1.84 KB |

AnthonySteele:
I think we should go back to `SendTo` for the time being. Can revisit that when we use Spans.

AnthonySteele:
I have the benchmark code on master and merged in here. Now that the stat names are the exact same length in both cases, the allocation is "1.84 KB" in both cases.

Issue #56, opened 2018-04-20
Do we need to have Culture?


AnthonySteele:
Do we need to have a `CultureInfo` object at all?

The `Culture` is on the `StatsDConfiguration` and defaults to `CultureInfo.InvariantCulture`
If you change it, it affects the way that metrics values are formatted for decimals and maybe other things.
But these metrics are not formatted for display, they are formatted to send to a server.

Does a statsd server really need culture-specific formatting? Under what circumstances is it a good idea to change it? If not, let's delete it, and then we can look at removing use of `string.Format` as an optimisation. As long as we format with a user-supplied  `CultureInfo` object, this will be hard to change. And it looks like yet more overdesign.

martincostello:
I'd always wondered why you'd want to do this as well...

AnthonySteele:
So we _usually_  leave the culture as `InvariantCulture`. We know that our stats servers accepts that.
Do our stats servers accept other formats? We don't know, but can find out.
Do other people have stats servers that _require_ other culture settings? We don't know.
[There are some docs here](https://github.com/etsy/statsd/blob/master/docs/metric_types.md) but not much mention of floats or float formats.

slang25:
My bet is it was left in because they may have been unsure when first creating this library. Logically I can't see how it would be valid to use anything other than invariant.

AnthonySteele:
So If I'm sending a stat, e.g an increment, it's a UDP packet containing a line of text like `some.stat.name:1|c`
Can also be a float, I think? `some.stat.name:2.4|c`

The basic question here is, what would happen if [a different culture format](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html) was applied to that float, e.g. 
`some.stat.name:2,4|c`

martincostello:
Try it with something _really_ different like Japanese etc. where it doesn't even use the same number characters and see what happens. I'm starting to think more and more this is a case of YAGNI.

AnthonySteele:
It was most likely introduced to prevent this code from picking up the server or UI culture format, which could easily be e.g. French or Danish. Making it configurable however looks like YAGNI overengineering, and is now preventing optimisations.

slang25:
I was nosing around other .NET implementations, they use `InvariantCulture` ¯\_(ツ)_/¯

martincostello:
Let's 🗑 it then.

justeattech:
We left it in becuase Resharper told us to...lol

[...]

PR #60, opened 2018-04-26
Culture info should not be configurable

AnthonySteele:

Simplify. We always want to use `CultureInfo.InvariantCulture` in the formatter. Do not allow this to be user configurable via the settings object.

#56

### DIFF
--- a/src/JustEat.StatsD/StatsDMessageFormatter.cs
+++ b/src/JustEat.StatsD/StatsDMessageFormatter.cs
@@ -12,15 +12,14 @@ namespace JustEat.StatsD
         [ThreadStatic]
         private static Random _random;
 
-        private readonly CultureInfo _cultureInfo;
+        private readonly CultureInfo _cultureInfo = CultureInfo.InvariantCulture;
         private readonly string _prefix;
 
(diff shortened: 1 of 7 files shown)


src/JustEat.StatsD/StatsDMessageFormatter.cs
```diff
@@ -12,15 +12,14 @@ public class StatsDMessageFormatter
         [ThreadStatic]
         private static Random _random;
 
-        private readonly CultureInfo _cultureInfo;
+        private readonly CultureInfo _cultureInfo = CultureInfo.InvariantCulture;
```
martincostello: This could be static now.

src/JustEat.StatsD/StatsDMessageFormatter.cs
```diff
@@ -12,15 +12,14 @@ public class StatsDMessageFormatter
         [ThreadStatic]
         private static Random _random;
 
-        private readonly CultureInfo _cultureInfo;
+        private static readonly CultureInfo CultureInfo = CultureInfo.InvariantCulture;
```
stuart-lang: Very minor comment, I'd name this `InvariantCulture`, or something to that effect

PR #62, opened 2018-04-28
Benchmark the formatter

AnthonySteele:

Benchmarks directly on `StatsDMessageFormatter`. 

### DIFF
--- a/src/Benchmark/FormatterBenchmark.cs
+++ b/src/Benchmark/FormatterBenchmark.cs
@@ -15,7 +15,13 @@ namespace Benchmark
         }
 
         [Benchmark]
-        public string Increment()
+        public string IncrementBy1()
+        {
+            return _formatter.Increment(1, "some.stat");
+        }
+
+        [Benchmark]
+        public string IncrementBy12()
         {
             return _formatter.Increment(12, "some.stat");
         }
--- a/src/Benchmark/FormatterBenchmark.cs
+++ b/src/Benchmark/FormatterBenchmark.cs
@@ -6,13 +6,7 @@ namespace Benchmark
     [MemoryDiagnoser]
     public class FormatterBenchmark
     {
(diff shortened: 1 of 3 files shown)


src/Benchmark/FormatterBenchmark.cs
```diff
@@ -0,0 +1,54 @@
...
+using JustEat.StatsD;
+
+namespace Benchmark
+{
+    [MemoryDiagnoser]
+    public class FormatterBenchmark
+    {
+        private StatsDMessageFormatter _formatter;
+
+        [GlobalSetup]
```
martincostello: Does this need to be a setup method? Or can it just be in the constructor?

src/Benchmark/FormatterBenchmark.cs
```diff
@@ -0,0 +1,54 @@
...
+        public void Setup()
+        {
+            _formatter = new StatsDMessageFormatter("test.prefix");
+        }
+
+        [Benchmark]
+        public string IncrementBy1()
+        {
+            return _formatter.Increment(1, "some.stat");
```
martincostello: Could all these methods just be `void` given that BenchmarkDotNet throws away the return values anyway?

AnthonySteele: IIRC, BenchmarkDotNet  docs say to return a value so that code won't be optimised out.

martincostello: I guess it depends on the case? The compiler couldn’t optimise everything away in case there was a side effect. Always returning probably removes the need to worry about whether it will or not.

AnthonySteele: I read it here http://benchmarkdotnet.org/Guides/GoodPractices.htm

AnthonySteele:
I would rather get this PR merged first, since it measures the other "fast path" one.

PR #68, opened 2018-05-04
Build version infrastructure

AnthonySteele:

Using scripts and patterns from NLog.StructuredLogging.Json and JustSaying to set version at time of build.


JustEat.StatsD.Common.props
```diff
@@ -1,4 +1,5 @@
 <Project>
+ <Import Project="version.props" />
```
martincostello: While you're at it, if you rename this to `Directory.Build.props`, it won't need manually importing by name in the `csproj` files.

AnthonySteele: Done

PR #70, opened 2018-05-17
metadata for 3.0.0 Beta 02

AnthonySteele:

metadata for 3.0.0 Beta 02 which will be used to test last changes to add `Timer` overloads.


martincostello:
Would be good to come up with a way to automate this at some point.

PR #71, opened 2018-05-26
Use SourceLink

martincostello:
* Add support for [SourceLink](https://github.com/dotnet/sourcelink) to improve the debugging experience in consuming applications (suggested in #69).
  * Update to .NET Core SDK `2.1.300` (needed for SourceLink support).
  * Update the .NET Core test SDK and Moq to the latest stable versions.
  * Move install steps in Travis from the YAML file to `build.sh`.
  * Read the SDK version from `global.json` so it only needs updating in one place going forwards.
  * Remove redundant `dotnet restore` step.
  * Ignore `BenchmarkDotNet.Artifacts` anywhere in the source tree.

Hold off on merging in case the SDK is updated again before the planned release date on the 30th May ([ref](https://github.com/aspnet/Home/wiki/2.1.0-Early-Access-Downloads#is-this-really-the-final-build)).


martincostello:
The magic part is that this line is now baked into the `.nuspec` in the NuGet package:

```xml
<repository type="git" url="https://github.com/justeat/JustEat.StatsD.git" commit="95ce41b75d401c41bce97467e2369a8dc7c3c2f9" />
```

AnthonySteele:
Looks good to me, but I don't know all that much about Sourcelink.

An observation is that these are some unrelated changes (e.g.  updating test SDK and Moq has nothing to do with Sourcelink and the test Benchmark.Net artefacts are again unrelated)

But there is a general theme of preparing for the NETCore 2.1 release, and a subsequent StatsD 3.0 release.

We are using the beta in 1 or 2 prod features, and so far it's fine. I'm happy to prepare all systems for launch. 🚀

martincostello:
Yeah they’re unrelated, but that’s easier to do in one PR as separate commits rather than raise 4 PRs and merge them all back together later.

`2.1.300` is a requirement for SourceLink to actually work, and the other two things I discovered while I was in the `csproj` files doing the SourceLink changes.

PR #73, opened 2018-06-18
Target NetStandard 2.0

AnthonySteele:

Target `NetStandard 2.0` as this target is very useful to have.
Leave existing platform targets  `net451`

Remove `NetStandard1.6`, as per comment below.

> we really want everyone to re-baseline at  netstandard 2.0. At the very least add a netstandard 2.0 target
 
David Fowler

https://twitter.com/davidfowl/status/1008453902058917888

### DIFF
--- a/Build.ps1
+++ b/Build.ps1
@@ -100,7 +100,7 @@ if ($RestorePackages -eq $true) {
 
 Write-Host "Building $($projects.Count) projects..." -ForegroundColor Green
 ForEach ($project in $projects) {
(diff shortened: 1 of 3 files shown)


AnthonySteele:
looking at it now, what does the Netstandard 1.6 target get us? Every app that I can think of is covered by either the `NetStandard 2.0` target or the `net451` target. We only need those two? And the `net451` is a minority.

martincostello:
@AnthonySteele Looks like the build scripts need a tweak.

PR #75, opened 2018-06-24
Delete unused classes

AnthonySteele:

Delete unused classes `PoolAwareSocketAsyncEventArgs` and `SimpleObjectPool`
We stopped using these in the last release, but the code is still there


AnthonySteele:
Check away. They are `public` :( but I doubt it. 

In the unlikely case that someone is, they should copy the source into their  app.

martincostello:
Yeah, I do use these. I use them in a alternate `IStatsDTransport` implementation to pool `UdpClient` instances.

AnthonySteele:
So is that use just in one place? would you rather have these classes in `JustEat.StatsD` or with the alternate `IStatsDTransport` implementation?

martincostello:
I guess it's easy enough for me to copy them over at whatever point I upgrade to StatsD 3, but I thought I'd point it out as it's another breaking change.

martincostello:
See #81, where we _might_ want to at least use `SimpleObjectPool`.

AnthonySteele:
Will review this later. We might make them internal and/or sealed

PR #76, opened 2018-07-08
Do not create 800b socket if we have nothing to send

dv00d00:
There is a case on StatsDMessageFormatter when Random.NextDouble() is less than a sample rate. The formatted result is an empty string in this case. There is no value in creating a socket for sending empty payload.


AnthonySteele:
Do the same in `UdpTransport`

prefered format is
`````
if (nothingToDo)
{
  return;
}
`````

martincostello:
Do we have any test cases anywhere for this? They should be safe to run in CI given they're no-ops.

PR #77, opened 2018-07-08
[RFC] Spans

dv00d00:
This is a draft of string formatting with spans. Benchamrk.Net shows 0b allocated for formatting and constant 800b for transporting stats. 

Is there an option for such kind of change to get merged?

### DIFF
--- a/src/Benchmark/Benchmark.csproj
+++ b/src/Benchmark/Benchmark.csproj
@@ -2,11 +2,19 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
 
   <PropertyGroup>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Benchmark/FormatterBenchmark.cs
+++ b/src/Benchmark/FormatterBenchmark.cs
@@ -1,3 +1,4 @@
+using System;
 using BenchmarkDotNet.Attributes;
 using JustEat.StatsD;
 
@@ -42,6 +43,60 @@ namespace Benchmark
         public string Timing()
         {
             return _formatter.Timing(1000, "some.stat");
+        }
+    }
+
+    [MemoryDiagnoser]
+    public class FormatterBenchmarkSpan
+    {
+        private readonly SpanStatsDMessageFormatter _formatter = new SpanStatsDMessageFormatter("test.prefix");
+
+        [Benchmark]
+        public void IncrementBy1()
+        {
+            Span<byte> buff = stackalloc byte[64];
+            var writer = new Writer(buff);
+            _formatter.Increment(1, "some.stat", ref writer);
+        }
+
+        [Benchmark]
+        public void IncrementBy12()
+        {
+            Span<byte> buff = stackalloc byte[64];
+            var writer = new Writer(buff);
+            _formatter.Increment(12, "some.stat", ref writer);
+        }
+
+        [Benchmark]
+        public void Decrement()
+        {
+            Span<byte> buff = stackalloc byte[64];
+            var writer = new Writer(buff);
+            _formatter.Decrement(12, "some.stat", ref writer);
+        }
+
+        [Benchmark]
+        public void Event()
+        {
+            Span<byte> buff = stackalloc byte[64];
+            var writer = new Writer(buff);
+            _formatter.Event("some.stat", ref writer);
+        }
+
+        [Benchmark]
+        public void Gauge()
+        {
+            Span<byte> buff = stackalloc byte[64];
+            var writer = new Writer(buff);
+            _formatter.Gauge(12, "some.stat", ref writer);
+        }
+
+        [Benchmark]
+        public void Timing()
+        {
+            Span<byte> buff = stackalloc byte[64];
+            var writer = new Writer(buff);
+            _formatter.Timing(1000, "some.stat", ref writer);
         }
     }
 
--- a/src/Benchmark/FormatterBenchmark.cs
+++ b/src/Benchmark/FormatterBenchmark.cs
@@ -55,7 +55,7 @@ namespace Benchmark
         public void IncrementBy1()
         {
             Span<byte> buff = stackalloc byte[64];
-            var writer = new Writer(buff);
+            var writer = new FixedBuffer(buff);
             _formatter.Increment(1, "some.stat", ref writer);
         }
 
@@ -63,7 +63,7 @@ namespace Benchmark
         public void IncrementBy12()
         {
             Span<byte> buff = stackalloc byte[64];
-            var writer = new Writer(buff);
+            var writer = new FixedBuffer(buff);
             _formatter.Increment(12, "some.stat", ref writer);
         }
 
@@ -71,7 +71,7 @@ namespace Benchmark
         public void Decrement()
         {
             Span<byte> buff = stackalloc byte[64];
-            var writer = new Writer(buff);
+            var writer = new FixedBuffer(buff);
             _formatter.Decrement(12, "some.stat", ref writer);
         }
 
@@ -79,7 +79,7 @@ namespace Benchmark
         public void Event()
         {
             Span<byte> buff = stackalloc byte[64];
-            var writer = new Writer(buff);
+            var writer = new FixedBuffer(buff);
             _formatter.Event("some.stat", ref writer);
         }
 
@@ -87,7 +87,7 @@ namespace Benchmark
         public void Gauge()
         {
             Span<byte> buff = stackalloc byte[64];
-            var writer = new Writer(buff);
+            var writer = new FixedBuffer(buff);
             _formatter.Gauge(12, "some.stat", ref writer);
         }
 
@@ -95,7 +95,7 @@ namespace Benchmark
         public void Timing()
         {
             Span<byte> buff = stackalloc byte[64];
-            var writer = new Writer(buff);
+            var writer = new FixedBuffer(buff);
             _formatter.Timing(1000, "some.stat", ref writer);
         }
     }
--- a/src/Benchmark/Program.cs
+++ b/src/Benchmark/Program.cs
@@ -26,8 +26,8 @@ namespace Benchmark
         static void Main(string[] args)
         {
             //BenchmarkRunner.Run<StatSendingBenchmark>(new FastAndDirty());
-            //BenchmarkRunner.Run<FormatterBenchmark>(new FastAndDirty());
-            BenchmarkRunner.Run<StatSendingBenchmark>(new FastAndDirty());
+            //BenchmarkRunner.Run<FormatterBenchmarkSpan>(new FastAndDirty());
+            BenchmarkRunner.Run<StatSendingBenchmark>();
         }
     }
 }
--- a/src/Benchmark/Program.cs
+++ b/src/Benchmark/Program.cs
@@ -28,16 +28,7 @@ namespace Benchmark
             var statSendingBenchmark = new StatSendingBenchmark();
             statSendingBenchmark.Setup();
 
-            //statSendingBenchmark.RunIp();
-
-            //for (int i = 0; i < 1000_000; i++)
-            //{
-            //    statSendingBenchmark.RunUdpSpan();
-            //}
-            
-
             BenchmarkRunner.Run<StatSendingBenchmark>(new FastAndDirty());
-            //BenchmarkRunner.Run<StatSendingBenchmark>();
         }
     }
 }
--- a/src/Benchmark/Program.cs
+++ b/src/Benchmark/Program.cs
@@ -25,9 +25,19 @@ namespace Benchmark
     {
         static void Main(string[] args)
         {
-            //BenchmarkRunner.Run<StatSendingBenchmark>(new FastAndDirty());
-            //BenchmarkRunner.Run<FormatterBenchmarkSpan>(new FastAndDirty());
-            BenchmarkRunner.Run<StatSendingBenchmark>();
+            var statSendingBenchmark = new StatSendingBenchmark();
+            statSendingBenchmark.Setup();
+
+            //statSendingBenchmark.RunIp();
+
+            //for (int i = 0; i < 1000_000; i++)
+            //{
+            //    statSendingBenchmark.RunUdpSpan();
+            //}
+            
+
+            BenchmarkRunner.Run<StatSendingBenchmark>(new FastAndDirty());
+            //BenchmarkRunner.Run<StatSendingBenchmark>();
         }
     }
 }
--- a/src/Benchmark/Program.cs
+++ b/src/Benchmark/Program.cs
@@ -1,14 +1,40 @@
+using System.Linq;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Diagnosers;
+using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Running;
+using RunMode = BenchmarkDotNet.Jobs.RunMode;
 
 namespace Benchmark
 {
+    public class FastAndDirty : ManualConfig
+    {
+        public FastAndDirty()
+        {
+            var job = new Job("", RunMode.Short, InfrastructureMode.InProcess);
+            this.Add(job);
+            this.Add(DefaultConfig.Instance.GetLoggers().ToArray());
+            this.Add(DefaultConfig.Instance.GetColumnProviders().ToArray());
+            this.Add(DefaultConfig.Instance.GetAnalysers().ToArray());
+            this.Add(DefaultConfig.Instance.GetDiagnosers().ToArray());
+            this.Add(new MemoryDiagnoser());
+        }
+    }
 
     class Program
     {
         static void Main(string[] args)
         {
-            BenchmarkRunner.Run<StatSendingBenchmark>();
-            BenchmarkRunner.Run<FormatterBenchmark>();
+
+            //Span<byte> buff = stackalloc byte[128];
+            //var formatter = new SpanStatsDMessageFormatter("some.cool.prefix");
+            //var writer = new Writer(buff);
+            //formatter.Gauge(1000, "some.stat", ref writer);
+            //Console.WriteLine(writer.Show());
+
+            //BenchmarkRunner.Run<StatSendingBenchmark>(new FastAndDirty());
+            //BenchmarkRunner.Run<FormatterBenchmark>(new FastAndDirty());
+            BenchmarkRunner.Run<FormatterBenchmarkSpan>(new FastAndDirty());
         }
     }
 }
--- a/src/Benchmark/Program.cs
+++ b/src/Benchmark/Program.cs
@@ -25,16 +25,9 @@ namespace Benchmark
     {
         static void Main(string[] args)
         {
-
-            //Span<byte> buff = stackalloc byte[128];
-            //var formatter = new SpanStatsDMessageFormatter("some.cool.prefix");
-            //var writer = new Writer(buff);
-            //formatter.Gauge(1000, "some.stat", ref writer);
-            //Console.WriteLine(writer.Show());
-
             //BenchmarkRunner.Run<StatSendingBenchmark>(new FastAndDirty());
             //BenchmarkRunner.Run<FormatterBenchmark>(new FastAndDirty());
-            BenchmarkRunner.Run<FormatterBenchmarkSpan>(new FastAndDirty());
+            BenchmarkRunner.Run<StatSendingBenchmark>(new FastAndDirty());
         }
     }
 }
--- a/src/Benchmark/StatSendingBenchmark.cs
+++ b/src/Benchmark/StatSendingBenchmark.cs
@@ -34,7 +34,7 @@ namespace Benchmark
             _udpSender.Increment("startup.ud");
 
             _ipSender = new StatsDPublisher(config, ipTransport);
-            _udpSender.Increment("startup.ip");
+            _ipSender.Increment("startup.ip");
         }
 
         [Benchmark]
--- a/src/Benchmark/StatSendingBenchmark.cs
+++ b/src/Benchmark/StatSendingBenchmark.cs
@@ -1,15 +1,22 @@
 using System;
 using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Attributes.Columns;
+using BenchmarkDotNet.Configs;
 using JustEat.StatsD;
 using JustEat.StatsD.EndpointLookups;
 
 namespace Benchmark
 {
     [MemoryDiagnoser]
+    [GroupBenchmarksBy(BenchmarkLogicalGroupRule.ByCategory)]
+    [CategoriesColumn]
     public class StatSendingBenchmark
     {
         private IStatsDPublisher _udpSender;
         private IStatsDPublisher _ipSender;
+
+        private IStatsDPublisher _udpSenderSpan;
+        private IStatsDPublisher _ipSenderSpan;
 
         private static readonly TimeSpan Timed = TimeSpan.FromMinutes(1);
 
@@ -28,6 +35,7 @@ namespace Benchmark
                 config.Host, config.Port, config.DnsLookupInterval);
 
             var udpTransport = new UdpTransport(endpointSource);
+            var udpTransportPooled = new UdpTransportPooled(endpointSource);
             var ipTransport = new IpTransport(endpointSource);
 
             _udpSender = new StatsDPublisher(config, udpTransport);
@@ -35,20 +43,47 @@ namespace Benchmark
 
             _ipSender = new StatsDPublisher(config, ipTransport);
             _ipSender.Increment("startup.ip");
+
+            _udpSenderSpan = new SpanStatsDPublisher(config, udpTransportPooled);
+            _udpSenderSpan.Increment("startup.ud");
+
+            _ipSenderSpan = new SpanStatsDPublisher(config, ipTransport);
+            _ipSenderSpan.Increment("startup.ip");
         }
 
-        [Benchmark]
+        private static void Run(IStatsDPublisher sender)
+        {
+            sender.Increment(1, 0.5, "increment.stat");
+            sender.Decrement(1, 0.5, "increment.stat");
+            sender.Increment(1, "increment.stat");
+            sender.Decrement(1, "increment.stat");
+            sender.Timing(Timed, "timer.stat");
+            sender.Timing(Timed, 0.5, "timer.stat");
+            sender.Gauge(long.MinValue, "gauge.stat", DateTime.Now);
+        }
+
+        [Benchmark(Baseline = true),BenchmarkCategory("UDP")]
         public void RunUdp()
         {
-            _udpSender.Increment("increment.ud");
-            _udpSender.Timing(Timed, "timer.ud");
+            Run(_udpSender);
         }
 
-        [Benchmark]
+        [Benchmark, BenchmarkCategory("UDP")]
+        public void RunUdpSpanPooled()
+        {
+            Run(_udpSenderSpan);
+        }
+
+        [Benchmark(Baseline = true), BenchmarkCategory("TCP")]
         public void RunIp()
         {
-            _ipSender.Increment("increment.ip");
-            _ipSender.Timing(Timed, "timer.ip");
+            Run(_ipSender);
+        }
+
+        [Benchmark, BenchmarkCategory("TCP")]
+        public void RunIpSpan()
+        {
+            Run(_ipSenderSpan);
         }
     }
 }
--- a/src/JustEat.StatsD/DateTimeExtensions.cs
+++ b/src/JustEat.StatsD/DateTimeExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 namespace JustEat.StatsD
 {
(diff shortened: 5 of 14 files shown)


src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -47,87 +50,135 @@ public StatsDPublisher(StatsDConfiguration configuration)
...
+            Span<byte> buffer = stackalloc byte[512];
+            var writer = new FixedBuffer(buffer);
+            _spanFormatter.Timing(Convert.ToInt64(duration.TotalMilliseconds), bucket, ref writer);
+            Send(writer.Get());
         }
 
         public void Timing(TimeSpan duration, double sampleRate, string bucket)
         {
-            Send(_formatter.Timing(Convert.ToInt64(duration.TotalMilliseconds), sampleRate, bucket));
+            Span<byte> buffer = stackalloc byte[512];
```
slang25: Is this a reasonable hard limit, or does it need surfacing in some way?

src/JustEat.StatsD/IpTransport.cs
```diff
@@ -14,14 +14,26 @@ public IpTransport(IPEndPointSource endPointSource)
...
+            var endpoint = _endpointSource.GetEndpoint();
+            Socket socket = null;
+            byte[] buffer = null;
+            
+            try
+            {
+                buffer = ArrayPool<byte>.Shared.Rent(metric.Length);
+                metric.CopyTo(buffer);
+                socket = CreateSocket();
+                socket.SendTo(buffer, metric.Length, SocketFlags.None, endpoint);
```
slang25: There are `Span` overloads of [`Socket.Send`](https://apisof.net/catalog/System.Net.Sockets.Socket.Send(ReadOnlySpan%3CByte%3E,SocketFlags)), but I found they just allocate internally, but I wasn't convinced it wasn't be doing something dumb.

slang25:
### FormatterBenchmark
        Method |      Mean |      Error |    StdDev |  Gen 0 | Allocated |
-------------- |----------:|-----------:|----------:|-------:|----------:|
  IncrementBy1 |  20.75 ns |   3.471 ns | 0.1961 ns | 0.0127 |      80 B |
 IncrementBy12 |  79.92 ns |  25.707 ns | 1.4525 ns | 0.0279 |     176 B |
     Decrement |  89.07 ns |  59.937 ns | 3.3866 ns | 0.0279 |     176 B |
         Event |  21.25 ns |   8.006 ns | 0.4523 ns | 0.0127 |      80 B |
         Gauge | 186.38 ns | 161.779 ns | 9.1408 ns | 0.0165 |     104 B |
        Timing |  90.21 ns |  19.432 ns | 1.0979 ns | 0.0304 |     192 B |

### StatSendingBenchmark
 Method |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
------- |----------:|----------:|----------:|-------:|----------:|
 RunUdp | 103.32 us | 50.502 us | 2.8535 us | 0.1221 |     800 B |
  RunIp |  74.36 us |  6.203 us | 0.3505 us | 0.1221 |     800 B |

src/JustEat.StatsD/StatsDPublisher.cs
dv00d00: This is just a random constant I choose from the top of my head, we should be able to quickly get the approximate minimum amount of bytes needed based on the metric string length.

src/JustEat.StatsD/IpTransport.cs
dv00d00: Send and SendTo are handled by different winsock methods internally, I did not check any further. There is no overload for SendTo accepting span, but that might get fixed soon.

AnthonySteele:
What are the before and after metrics from benchmarkDotNet, on the same machine?

What are the platforms supported by this code? We currently target `net451` and `netstandard2.0`.  Though we could increment or drop the older platform, but only if we do a new major version.

dv00d00:
I kept the old formatter and benchamrks for it. Span based results are:

|Method |      Mean |     Error |     StdDev |    Median | Allocated |
|-------------- |----------:|----------:|-----------:|----------:|----------:|
|  IncrementBy1 |  99.53 ns | 4.2938 ns | 12.2505 ns |  93.66 ns |       0 B |
| IncrementBy12 | 135.96 ns | 2.7377 ns |  5.4674 ns | 135.36 ns |       0 B |
|     Decrement | 146.06 ns | 1.8600 ns |  1.6488 ns | 146.12 ns |       0 B |
|         Event |  88.29 ns | 1.1799 ns |  1.1037 ns |  88.12 ns |       0 B |
|         Gauge | 127.66 ns | 0.6875 ns |  0.5368 ns | 127.78 ns |       0 B |
|        Timing | 133.87 ns | 2.6460 ns |  2.9410 ns | 132.76 ns |       0 B |

dv00d00:
String-based formatter:

|Method |      Mean |     Error |   StdDev |    Median |  Gen 0 | Allocated |
|-------------- |----------:|----------:|---------:|----------:|-------:|----------:|
|  IncrementBy1 |  29.58 ns | 0.6804 ns | 1.118 ns |  29.26 ns | 0.0508 |      80 B |
| IncrementBy12 | 110.60 ns | 1.5780 ns | 1.318 ns | 110.43 ns | 0.1118 |     176 B |
|     Decrement | 120.23 ns | 2.0964 ns | 1.751 ns | 120.35 ns | 0.1118 |     176 B |
|         Event |  31.12 ns | 0.7074 ns | 1.775 ns |  30.39 ns | 0.0508 |      80 B |
|         Gauge | 247.25 ns | 4.9498 ns | 7.559 ns | 244.57 ns | 0.0658 |     104 B |
|        Timing | 126.62 ns | 2.0130 ns | 1.681 ns | 126.21 ns | 0.1218 |     192 B |

dv00d00:
@AnthonySteele to get full benefit out of it the proper choice would be netcoreapp2.1+, and probably .net 4.8 if it gets native spans.

It still might be worth doing on a slowspan implementations (.net and netcoreapp < 2.1). I am not entirely sure if span based apis will work there, I will investigate.

src/JustEat.StatsD/IpTransport.cs
slang25: Cool, I wasn't aware of that. I was under the impression the `SendTo` version was a convenience wrapper for the typical use case (see here: https://github.com/justeat/JustEat.StatsD/pull/55/commits/c18e070e5ee44cca6453b01528d04cb7b9babe1a)

AnthonySteele:
My main concern is that the span run seems to be slower. [And if the benchmark is using 2.1, it should use the fast spans?](https://github.com/justeat/JustEat.StatsD/pull/77/files#diff-64354ac04d2b52d7a64e6a8950aebe38R5)

dv00d00:
It looks like it is slower due to UTF8.GetBytes is called at formatter level, previously UTF8 get bytes was called before sending the packet over the wire

dv00d00:
Previous implementation benchmark on my machine: 

|Method |     Mean |    Error |   StdDev |   Median | Allocated |
|------- |---------:|---------:|---------:|---------:|----------:|
| RunUdp | 832.3 us | 19.88 us | 56.07 us | 815.3 us |   1.16 KB |
|  RunIp | 774.6 us | 15.19 us | 21.30 us | 766.6 us |   1.16 KB |

This:

|Method |     Mean |     Error |    StdDev | Allocated |
|------- |---------:|----------:|----------:|----------:|
| RunUdp | 783.8 us | 12.176 us | 10.167 us |     800 B |
|  RunIp | 766.3 us |  8.581 us |  6.700 us |     800 B |

dv00d00:
I will run benchmarks from my laptop, for some reason, it performs 8 times slower than a desktop.
After the latest change to transport with Connect-Send methods performance degraded approximately by 5%

It feels like major perf benefit would come from Socket's pooling, via something similar to HttpClientFactory but for sockets.

Perf boost from utf8 parser writing into stackalocked span is only slightly faster than String.Format, there might be some perf boost potential hidden there as well.

Reducing stacklocked buffer size to something dependant on payload helped a bit, will scratch it there as well.

martincostello:
A quick two-pence thought into this PR (though I don't have time to properly review it until tomorrow), I'm concerned that in the rush to jump on the new-shiny-thing of Spans, we'd seem to be forcing the library to only support the latest version of .NET Core/Standard, which would suddenly leave users with slower release trains a bit left out in the dark for bugs fixes etc.

slang25:
It could be `#if`ed to keep support for the current TFMs, or the other existing TFMs could take a dependency on `System.Memory`, which polyfill with "slow Span", which will still be crazy quick with this.

I vote for minimum impact to existing TFMs, so `#if` where needed.

dv00d00:
Several updates:

I've added a benchmark with empty transport implementations. In them, span based version runs 33 % faster. But the absolute time change is in a range of milliseconds. The really slow part is a socket creation and IO it produces.

Average time to create a socket and send a packet via it is approximately 100ms on my machine. Perf boost from using non-allocating formatting falls into a range of standard deviation of sending the actual data over the wire.

I will focus on separating new formatter from existing stuff to make it optional and keep compatibility with older versions of .net.

dv00d00:
So I've tried to add socket pooling for UDP transport

|           Method | Categories |      Mean |     Error |    StdDev | Scaled |  Gen 0 | Allocated |
|----------------- |----------- |----------:|----------:|----------:|-------:|-------:|----------:|
|           RunUdp |        UDP | 315.91 us | 49.740 us | 2.8104 us |   1.00 | 0.9766 |    4029 B |
| RunUdpSpanPooled |        UDP |  54.18 us |  6.273 us | 0.3544 us |   0.17 |      - |       0 B |
|                  |            |           |           |           |        |        |           |
|            RunIp |        TCP | 290.65 us | 22.373 us | 1.2641 us |   1.00 | 0.9766 |    4045 B |
|        RunIpSpan |        TCP | 284.13 us | 26.288 us | 1.4853 us |   0.98 | 0.4883 |    2208 B |

martincostello:
I’ll look at the results again tomorrow, but socket-overhead is something that’s come up internally. We have a couple of apps using their own socket-pooling implementation, but the CPU overhead wasn’t quite quantifiable enough in benchmarks to make the case for it being the default implementation here in the main library.

src/JustEat.StatsD/UdpTransportPooled.cs
```diff
@@ -0,0 +1,53 @@
...
+
+#if !NET451
+                // See https://github.com/dotnet/corefx/pull/17853#issuecomment-291371266
+                if (!RuntimeInformation.IsOSPlatform(OSPlatform.OSX)) socket.SendBufferSize = 0;
+#else
+                socket.SendBufferSize = 0;
+#endif
+
+                _socket?.Dispose();
+                _ipEndPoint = endPoint;
```
slang25: Nitpicking here possibly, this line and the next should be swapped so that there's no weirdness in case of `socket.Connect` throwing and the 2 fields being in a mixed state. This is really edge case stuff though

src/Benchmark/Benchmark.csproj
```diff
@@ -2,13 +2,21 @@
...
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
 
   <PropertyGroup>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <LangVersion>latest</LangVersion>
```
martincostello: Should just set this in the root `Directory.Build.props` and then it'll apply to all projects.

src/Benchmark/FormatterBenchmark.cs
```diff
@@ -45,4 +46,58 @@ public string Timing()
         }
     }
 
+    [MemoryDiagnoser]
+    public class FormatterBenchmarkSpan
```
martincostello: If we'd keep this going forwards, put this into its own file.

src/Benchmark/Program.cs
```diff
@@ -1,14 +1,34 @@
...
 using BenchmarkDotNet.Running;
+using RunMode = BenchmarkDotNet.Jobs.RunMode;
 
 namespace Benchmark
 {
+    public class FastAndDirty : ManualConfig
+    {
+        public FastAndDirty()
+        {
+            var job = new Job("", RunMode.Short, InfrastructureMode.InProcess);
```
martincostello: I _think_ this could be simplified to just this so you don't need to add all the loggers etc. in.

```csharp
public FastAndDirty()
{
    var job = Job.Core
        .With(InProcessToolchain.Instance)
        .WithId("InProcess");

    Add(job);
    Add(MemoryDiagnoser.Default);
}
```

src/Benchmark/Program.cs
```diff
@@ -1,14 +1,34 @@
...
+    }
 
     class Program
     {
         static void Main(string[] args)
         {
-            BenchmarkRunner.Run<StatSendingBenchmark>();
-            BenchmarkRunner.Run<FormatterBenchmark>();
+            var statSendingBenchmark = new StatSendingBenchmark();
+            statSendingBenchmark.Setup();
```
martincostello: Doesn't BenchmarkDotNet ensure this gets done itself already as part of `Run<>()`?

src/Benchmark/StatSendingBenchmark.cs
```diff
@@ -28,27 +35,55 @@ public void Setup()
...
+            sender.Decrement(1, 0.5, "increment.stat");
+            sender.Increment(1, "increment.stat");
+            sender.Decrement(1, "increment.stat");
+            sender.Timing(Timed, "timer.stat");
+            sender.Timing(Timed, 0.5, "timer.stat");
+            sender.Gauge(long.MinValue, "gauge.stat", DateTime.Now);
         }
 
-        [Benchmark]
+        [Benchmark(Baseline = true),BenchmarkCategory("UDP")]
```
martincostello: One line per attribute is neater, IMHO.

src/JustEat.StatsD/DateTimeExtensions.cs
```diff
@@ -12,9 +12,10 @@ public static class DateTimeExtensions
         /// <summary>	Returns the DateTime as the number of seconds since the epoch (1970), which is Unix time. </summary>
         /// <param name="dateTime">	The dateTime to act on. </param>
         /// <returns>	A number of seconds since the epoch. </returns>
-        public static double AsUnixTime(this DateTime dateTime)
+        public static long AsUnixTime(this in DateTime dateTime)
         {
-            return Math.Round(dateTime.ToUniversalTime().Subtract(Epoch).TotalSeconds);
+            DateTimeOffset x = dateTime;
```
martincostello: Relying on the implicit cast can cause issues with time zone conversions if `dateTime` isn't UTC.

src/JustEat.StatsD/IStatsDTransport.cs
```diff
@@ -1,7 +1,9 @@
+using System;
+
 namespace JustEat.StatsD
 {
     public interface IStatsDTransport
     {
-        void Send(string metric);
+        void Send(ReadOnlySpan<byte> metric);
```
martincostello: Would be a breaking change and prevent use by any TFM that isn't 2.1+ compatible.

src/JustEat.StatsD/IpTransport.cs
```diff
@@ -14,14 +14,16 @@ public IpTransport(IPEndPointSource endPointSource)
             _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
         }
 
-        public void Send(string metric)
+        public void Send(ReadOnlySpan<byte> metric)
         {
-            var endpoint = _endpointSource.GetEndpoint();
-            var bytes = Encoding.UTF8.GetBytes(metric);
+            if (metric.Length == 0)
+                return;
```
martincostello: Braces:

```csharp
if (metric.Length == 0)
{
    return;
}
```

src/JustEat.StatsD/JustEat.StatsD.csproj
```diff
@@ -1,11 +1,17 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Description>A .NET library for publishing metrics to statsd.</Description>
     <OutputType>Library</OutputType>
     <PackageId>JustEat.StatsD</PackageId>
     <RootNamespace>JustEat.StatsD</RootNamespace>
     <Summary>A .NET library for publishing metrics to statsd.</Summary>
-    <TargetFrameworks>net451;netstandard2.0</TargetFrameworks>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
martincostello: If we're a generic library, we shouldn't be targeting `netcoreappx.x`

src/JustEat.StatsD/JustEat.StatsD.csproj
```diff
@@ -15,4 +21,7 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.0.0" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="1.1.0" />
```
martincostello: Why the additional reference for all TFMs of `1.1.0` when there's a `2.0.0` on the line above that would be enabled if the `Condition` was removed?

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
```diff
@@ -0,0 +1,266 @@
...
+namespace JustEat.StatsD
+{
+    public static class FixedBufferExtensions
+    {
+        public static string Show(in this FixedBuffer src) => Encoding.UTF8.GetString(src.Written());
+
+        public static Span<byte> Written(in this FixedBuffer src) => src.Buffer.Slice(0, src.Position);
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ref FixedBuffer Add(ref this FixedBuffer src, byte[] str)
```
martincostello: `str` isn't a great name for a variable that's a byte array.

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
```diff
@@ -0,0 +1,266 @@
+using System;
+using System.Buffers;
+using System.Buffers.Text;
+using System.Globalization;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace JustEat.StatsD
+{
+    public static class FixedBufferExtensions
```
martincostello: Does this need to be `public`?

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
martincostello: `source` not `src` (or `buffer`?).

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
```diff
@@ -0,0 +1,266 @@
...
+            return ref src;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ref FixedBuffer Add(ref this FixedBuffer src, double value)
+        {
+            Span<char> dest = stackalloc char[32];
+            value.TryFormat(dest, out int written, "", CultureInfo.InvariantCulture);
+
+
```
martincostello: Too much whitespace.

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
```diff
@@ -0,0 +1,266 @@
...
+            var written = Encoding.UTF8.GetBytes(str, src.Buffer.Slice(src.Position));
+            src.Position += written;
+            return ref src;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ref FixedBuffer Add(ref this FixedBuffer src, double value)
+        {
+            Span<char> dest = stackalloc char[32];
+            value.TryFormat(dest, out int written, "", CultureInfo.InvariantCulture);
```
martincostello: Use `string.Empty`.

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            var written = Encoding.UTF8.GetBytes(str, src.Buffer.Slice(src.Position));
+            src.Position += written;
+            return ref src;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ref FixedBuffer Add(ref this FixedBuffer src, double value)
+        {
+            Span<char> dest = stackalloc char[32];
```
martincostello: Why `32`?

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
```diff
@@ -0,0 +1,266 @@
...
+            Buffer = buffer;
+            Position = 0;
+        }
+    }
+
+    public class SpanStatsDMessageFormatter
+    {
+        private const double DefaultSampleRate = 1.0;
+
+        [ThreadStatic]
```
martincostello: Why would we need a `Random` per thread?

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
```diff
@@ -0,0 +1,266 @@
...
+        public SpanStatsDMessageFormatter(string prefix)
+        {
+            _prefix = !string.IsNullOrWhiteSpace(prefix) ?
+                Encoding.UTF8.GetBytes(prefix + ".") :
+                Array.Empty<byte>();
+        }
+
+        private static Random Random => _random ?? (_random = new Random());
+
+        public void Timing(long milliseconds, string statBucket, ref FixedBuffer fixedBuffer)
```
martincostello: I'm not sure we'd want to create a new public type with all these "advanced" overloads using a custom struct.

src/JustEat.StatsD/SpanStatsDPublisher.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            var endpointSource = EndpointParser.MakeEndPointSource(
+                configuration.Host, configuration.Port, configuration.DnsLookupInterval);
+            _transport = new UdpTransport(endpointSource);
+            _onError = configuration.OnError;
+        }
+
+        public void Increment(string bucket)
+        {
+            Span<byte> buffer = stackalloc byte[_spanFormatter.GetMaxSize(bucket)];
```
martincostello: These two lines are repeated again and again in all the methods. Is there a way to use `ref return` or something in a helper method to reduce the copy-paste?

src/JustEat.StatsD/SpanStatsDPublisher.cs
```diff
@@ -0,0 +1,196 @@
...
+            Send(writer.Written());
+        }
+
+        public void Increment(long value, double sampleRate, string bucket)
+        {
+            Span<byte> buffer = stackalloc byte[_spanFormatter.GetMaxSize(bucket)];
+            var writer = new FixedBuffer(buffer);
+            _spanFormatter.Increment(value, sampleRate, bucket, ref writer);
+            Send(writer.Written());
+
```
martincostello: Unneeded whitespace.

src/JustEat.StatsD/StatsDMessageFormatter.cs
```diff
@@ -5,6 +5,57 @@
 
 namespace JustEat.StatsD
 {
+    internal static class StringBuilderCache
+    {
+        private const int DefaultCapacity = 256;
+        private const int MaxBuilderSize = DefaultCapacity * 3;
+
+        [ThreadStatic]
```
martincostello: Is this just being used as a way to remove the need to worry about locks?

src/JustEat.StatsD/StatsDMessageFormatter.cs
```diff
@@ -5,6 +5,57 @@
...
+        private static StringBuilder t_cachedInstance;
+
+        /// <summary>Written a StringBuilder for the specified capacity.</summary>
```
martincostello: This description isn't clear.

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -9,15 +10,12 @@ namespace JustEat.StatsD
     public class StatsDPublisher : IStatsDPublisher
     {
         private readonly StatsDMessageFormatter _formatter;
-        private readonly IStatsDTransport _transport;
         private readonly Func<Exception, bool> _onError;
+        private readonly IStatsDTransport _transport;
 
         public StatsDPublisher(StatsDConfiguration configuration, IStatsDTransport transport)
         {
-            if (configuration == null)
```
martincostello: Put this back as it was.

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -131,15 +127,12 @@ private void Send(string metric)
         {
             try
             {
-                _transport.Send(metric);
+                _transport.Send(Encoding.UTF8.GetBytes(metric));
```
martincostello: This is the only change that should be committed from this file.

src/JustEat.StatsD/UdpTransport.cs
```diff
@@ -17,14 +17,16 @@ public UdpTransport(IPEndPointSource endPointSource)
             _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
         }
 
-        public void Send(string metric)
+        public void Send(ReadOnlySpan<byte> metric)
         {
-            var bytes = Encoding.UTF8.GetBytes(metric);
-            var endpoint = _endpointSource.GetEndpoint();
+            if (metric.Length == 0)
```
martincostello: Braces.

src/JustEat.StatsD/UdpTransportPooled.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Runtime.InteropServices;
+using JustEat.StatsD.EndpointLookups;
+
+namespace JustEat.StatsD
+{
+    public class UdpTransportPooled : IStatsDTransport
+    {
+        private readonly IPEndPointSource _endpointSource;
+
+        [ThreadStatic] private static Socket _socket;
```
martincostello: This would make this class very difficult to test in isolation.

martincostello: Attributes should be on their own line.

src/JustEat.StatsD/UdpTransportPooled.cs
```diff
@@ -0,0 +1,53 @@
...
+        [ThreadStatic] private static IPEndPoint _ipEndPoint;
+
+        public UdpTransportPooled(IPEndPointSource endPointSource)
+        {
+            _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
+        }
+
+        public void Send(ReadOnlySpan<byte> metric)
+        {
+            if (metric.Length == 0)
```
martincostello: Braces.

src/JustEat.StatsD/UdpTransportPooled.cs
```diff
@@ -0,0 +1,53 @@
...
+            socket.Send(metric);
+        }
+
+        private static Socket GetSocket(IPEndPoint endPoint)
+        {
+            if (_socket == null || !ReferenceEquals(_ipEndPoint, endPoint))
+            {
+                var socket = new Socket(AddressFamily.InterNetwork, SocketType.Dgram, ProtocolType.Udp);
+
+#if !NET451
```
martincostello: With the TFMs as they are in this PR, this is redundant.

AnthonySteele:
I agree with Martin: the Spans have a lot of complexity, only run on a few of the frameworks that we use, produce small gains on something that's not the bottleneck.

AnthonySteele:
We had pooled resources, but it wasn't helping performance. See https://github.com/justeat/JustEat.StatsD/pull/75/files
Possibly we weren't doing it well or in the right place and it can be re-introduced with more modern platform support?

src/JustEat.StatsD/JustEat.StatsD.csproj
slang25: This is the nature of the situation that has come about.
There is lots of cool perf stuff currently exclusive to .NET Core 2.1, and not other runtimes.
For example, today I think `Span<T>` wont be shipping with .NET 4.8 (see comments here: https://blogs.msdn.microsoft.com/dotnet/2018/06/06/announcing-net-framework-4-8-early-access-build-3621/).
We'll probably need to wait for .NET Standard 3.0, which could be a very long way off.

IMO, for this PR to be more acceptable, this needs to be an additional target, with no breaking changes or additional deps for existing targets.

martincostello: But then how long would we support that for in future versions? From what I've seen in conversations online from the .NET guys like Immo and David Fowler, the guidance seems to be that if you're a wide-ranging general library, you should wait for other implementations to catch-up (e.g. Mono) and for the standard to be updated, rather than rush into supporting a specific implementation to get access to the new shiny things ASAP.

src/JustEat.StatsD/SpanStatsDMessageFormatter.cs
slang25: Concurrent access to `Random` is unsafe and can gives weird results, I've seen it return lot's of `0`s under high load.

martincostello: Fair enough - I've never heard of that problem before.

src/JustEat.StatsD/JustEat.StatsD.csproj
slang25: I've heard it more like "if you want the stuff we shipped in .NET Core 2.1, additionally target `netcoreapp2.1`, or wait until it hits the standard".

If done in a way that isn't detrimental to current consumers of this package, I think this PR is a great feature, and something to boast about.

slang25:
@martincostello https://youtu.be/LCV-JNOkZC4?t=6m5s
This clip pretty much covers my understanding. It's worth watching 10 minutes from there, my comment gets read out and your tweet and gets shown too.

martincostello:
I'd be much happier with a separate package until it's more reasonable to bring up the drawbridge on older TFMs, but even if we add an extra TFM and use `#if` now, we're still creating divergence as you need to make a breaking change to the core interface (for the 3rd time this year?) so you can pass in a span.

dv00d00:
Closing it cause it definitely would not be merged, will still refer to in future PRs.

PR #80, opened 2018-07-16
Shared socket

AnthonySteele:

Re-use a socket. This is not a user-visible breaking change, so could be a patch version.
However we need to find out if the socket is thread-safe, if there are any cases where a new socket is needs and how to do that in a thread-safe manner, and how to dispose of the socket.

Why would we do this? Check the before and after benchmarks:

Before:

|             Method |      Mean |    Error |   StdDev |  Gen 0 | Allocated |
|------------------- |----------:|---------:|---------:|-------:|----------:|
|              RunIp | 372.77 us | 4.851 us | 3.788 us |      - |    1184 B |
|  RunIpWithSampling |  96.21 us | 1.765 us | 1.651 us | 0.1221 |     506 B |

After:

|             Method |       Mean |     Error |    StdDev |  Gen 0 | Allocated |
------------------- |-----------:|----------:|----------:|-------:|----------:|
|              RunIp |  16.352 us | 0.0783 us | 0.0611 us | 0.2136 |     720 B |
|  RunIpWithSampling |   5.224 us | 0.0745 us | 0.0697 us | 0.1221 |     388 B |

### DIFF
--- a/src/JustEat.StatsD/IpTransport.cs
+++ b/src/JustEat.StatsD/IpTransport.cs
@@ -9,6 +9,8 @@ namespace JustEat.StatsD
     {
         private readonly IPEndPointSource _endpointSource;
 
+        private readonly Socket _socket = CreateSocket();
+
         public IpTransport(IPEndPointSource endPointSource)
         {
             _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
@@ -24,10 +26,7 @@ namespace JustEat.StatsD
             var endpoint = _endpointSource.GetEndpoint();
             var bytes = Encoding.UTF8.GetBytes(metric);
 
-            using (var socket = CreateSocket())
-            {
-                socket.SendTo(bytes, endpoint);
-            }
+            _socket.SendTo(bytes, endpoint);
         }
 
         private static Socket CreateSocket()


jmahmud:
@AnthonySteele - does a mechanism for re-creating a socket be part of this implementation - catching of exception / on certain conditions?  Not sure what those would be however...

AnthonySteele:
Yes, throwing the socket away if there's an error is possible, but we first need to know if `  socket.SendTo(bytes, endpoint)` is completely thread-safe and re-entrant, like `HttpClient.SendAsync`.

martincostello:
Maybe worth asking the question in the corefx (or whatever) repo? Alternatively, maybe have a cursory dig through the new `HttpClient` stack that's built on top of sockets and see what they do with them.

AnthonySteele:
[Source](https://github.com/dotnet/corefx/blob/master/src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs)

martincostello:
Love a good 5,000 line source file...

dv00d00:
Might be worth consideration:

|Method |     Mean |    Error |   StdDev | Scaled | ScaledSD | Allocated |
|------- |---------:|---------:|---------:|-------:|---------:|----------:|
| SendTo | 64.73 us | 179.1 us | 10.12 us |   1.00 |     0.00 |     168 B |
|   Send | 51.24 us | 255.5 us | 14.44 us |   0.80 |     0.21 |       0 B |


```csharp
    [MemoryDiagnoser]
    public class SocketTests
    {
        private Socket socket;
        private IPEndPoint ipEndPoint;
        private byte[] bytes;

        [GlobalSetup(Target = nameof(SendTo))]
        public void Setup()
        {
            socket = new Socket(AddressFamily.InterNetwork, SocketType.Dgram, ProtocolType.Udp);
            ipEndPoint = new IPEndPoint(new IPAddress(new byte[] { 127, 0, 0, 1 }), 8888);
            bytes = Encoding.ASCII.GetBytes("hello world yo");
        }

        [GlobalSetup(Target = nameof(Send))]
        public void SetupSend()
        {
            socket = new Socket(AddressFamily.InterNetwork, SocketType.Dgram, ProtocolType.Udp);
            ipEndPoint = new IPEndPoint(new IPAddress(new byte[] { 127, 0, 0, 1 }), 8888);
            bytes = Encoding.ASCII.GetBytes("hello world yo");
            socket.Connect(ipEndPoint);
        }

        [Benchmark(Baseline = true)]
        public void SendTo()
        {
            socket.SendTo(bytes, ipEndPoint);
        }

        [Benchmark]
        public void Send()
        {
            socket.Send(bytes);
        }
    }
```

dv00d00:
Also found this in socket handler for http client

https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/ConnectHelper.cs

might be useful. But they are dealing with tcp sockets, which are stateful, UDP socket might be much simpler to deal with.

AnthonySteele:
[Inquiry about sockets](https://github.com/dotnet/corefx/issues/29932)

AnthonySteele:
[No, sockets are not thread-safe](https://github.com/dotnet/corefx/issues/31117). If we do anything like this, it should be via a pool.

martincostello:
Here's the pooling implementation I've been using internally, based on the `2.1.0` codebase:

```csharp
using System;
using System.Collections.Generic;
using System.Linq;
using System.Net.Sockets;
using System.Runtime.InteropServices;
using JustEat.StatsD.EndpointLookups;

namespace JustEat.StatsD
{
    /// <summary>
    /// A class representing an implementation of <see cref="IStatsDTransport"/> that pools UDP clients. This class cannot be inherited.
    /// </summary>
    {
        private const int PoolSize = 30;

        private static readonly SimpleObjectPool<SocketAsyncEventArgs> EventArgsPool
            = new SimpleObjectPool<SocketAsyncEventArgs>(PoolSize, pool => new PoolAwareSocketAsyncEventArgs(pool));

        private readonly SimpleObjectPool<UdpClient> _clientPool
            = new SimpleObjectPool<UdpClient>(PoolSize, pool => CreateUdpClient());

        private readonly IPEndPointSource _endpointSource;

        /// <summary>
        /// Initializes a new instance of the <see cref="PooledUdpTransport"/> class.
        /// </summary>
        /// <param name="endPointSource">The <see cref="IPEndPointSource"/> to use.</param>
        /// <paramref name="endPointSource"/> is <see langword="null"/>.
        /// </exception>
        public PooledUdpTransport(IPEndPointSource endPointSource)
        {
        }

        /// <inheritdoc />
        public void Dispose()
        {
            UdpClient client = null;

            while ((client = _clientPool.Pop()) != null)
            {
                client.Dispose();
            }
        }

        /// <inheritdoc />
        public void Send(string metric)
        {
            Send(new[] { metric });
        }

        /// <inheritdoc />
        public void Send(IEnumerable<string> metrics)
        {
            if (metrics == null)
            {
                return;
            }

            var data = EventArgsPool.Pop();

            if (data == null)
            {
                return;
            }

            var udpClient = _clientPool.Pop();

            if (udpClient == null)
            {
                return;
            }

            try
            {

                bool connect = !udpClient.Client.Connected;

                // Is the endpoint unassigned or has changed?
                if (!endpoint.Equals(data.RemoteEndPoint))
                {
                    data.RemoteEndPoint = endpoint;
                    connect = true;
                }

                if (connect)
                {
                    udpClient.Client.Connect(data.RemoteEndPoint);
                }

                data.SendPacketsElements = metrics.ToMaximumBytePackets()
                    .Select(bytes => new SendPacketsElement(bytes, 0, bytes.Length, true))
                    .ToArray();

                udpClient.Client.SendPacketsAsync(data);
            }
            finally
            {
                _clientPool.Push(udpClient);
            }
        }

        private static UdpClient CreateUdpClient()
        {
            var client = new UdpClient();

            // See https://github.com/dotnet/corefx/pull/17853#issuecomment-291371266
            if (!RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
            {
                client.Client.SendBufferSize = 0;
            }

            return client;
        }
    }
}
```

martincostello:
I'll have a quick go at porting it to `3.0.0` now and do a PR for review.

martincostello:
PR: #81

AnthonySteele:
Concept proved, but the sockets are not thread-safe; abandoned in favour of pools.

PR #81, opened 2018-07-18
Pool Sockets in UdpTransport

martincostello:
As discussed in #80, this PR reworks the `UdpTransport` class to pool instances of `Socket` using existing infrastructure inside the library to reduce the overhead of socket creation when sending metrics. Note that the pool size of `30` is arbitrary.

The real win for this should be in reduction of CPU usage, rather than raw speed.

The implementation is based on one used internally at Just Eat in two different .NET Core 2.1 applications that was cooked-up while investigating high CPU usage that _appeared_ to point to JustEat.StatsD, but was actually a red-herring caused by misconfigured logging.

That implementation (based on the `2.1.0` codebase) has been in use in production for over 4 months and is in the code path for all the requests [in my blog post](https://tech.just-eat.com/2018/06/14/aspnet-core-21-supercharging-our-applications/) about the performance improvements from .NET Core, so I'm confident the experience with the different implementation is _at least_ not worse overall.

On reflection, changing the two types to `internal` that we originally discussed could be deleted (can't remember which issue/PR) is still a breaking change, meaning this would have to be `4.0.0`, so I'll probably put them back to `public` before merging (if we do).

Benchmarks to follow.

PR #84, opened 2018-07-18
Pool is flexibly sized

AnthonySteele:

`SimpleObjectPool` does not have a fixed size any more. It grows as needed using the constructor function. As used for Sockets in `PooledUdpTransport`, it starts small (1 socket per logical CPU core, i.e. 8 on my machine)

This means that you cannot use `Pop` in the `Dispose` as it will just create new items. So a `Count` property is needed to tell this when to stop.

There should not be a `try...finally` arounds socket use, as a socket that throws exception should not be put back in the pool.

FYI, the `Parallel.For` test uses a pool size of 8 (on my machine, assume that it's using 1 thread per core).

If I was breaking backwards compatibility I would also do:

- change `Func<SimpleObjectPool<T>, T>` to just `Func<T>` - the first param is not used at all.
* change the constructor of `SimpleObjectPool` to have the size as an optional param at the end. It is acceptable to start with a size of zero. i.e. `public SimpleObjectPool(Func<T> constructor, int initialSize = 0)`
* Why is it a _generic_ `SimpleObjectPool<T>` in the StatsD library, when we only ever put `Socket` objects in it?

### DIFF
--- a/src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs
+++ b/src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs
@@ -42,7 +42,7 @@ namespace JustEat.StatsD
 
             using (var target = new PooledUdpTransport(endPointSource))
             {
-                Parallel.For(0, 10_000, (_) =>
+                Parallel.For(0, 10_000, _ =>
                 {
                     target.Send("mycustommetric");
                 });
--- a/src/JustEat.StatsD/PooledUdpTransport.cs
+++ b/src/JustEat.StatsD/PooledUdpTransport.cs
@@ -42,7 +42,16 @@ namespace JustEat.StatsD
             var endpoint = _endpointSource.GetEndpoint();
 
             var socket = _socketPool.Pop();
-            socket.SendTo(bytes, endpoint);
+            try
+            {
+                socket.SendTo(bytes, endpoint);
+            }
+            catch (Exception)
+            {
+                socket.Dispose();
+                throw;
+            }
+
             _socketPool.Push(socket);
         }
 
--- a/src/JustEat.StatsD/PooledUdpTransport.cs
+++ b/src/JustEat.StatsD/PooledUdpTransport.cs
@@ -41,7 +41,7 @@ namespace JustEat.StatsD
             var bytes = Encoding.UTF8.GetBytes(metric);
             var endpoint = _endpointSource.GetEndpoint();
 
-            var socket = _socketPool.Pop();
+            var socket = _socketPool.PopOrCreate();
             try
             {
                 socket.SendTo(bytes, endpoint);
--- a/src/JustEat.StatsD/PooledUdpTransport.cs
+++ b/src/JustEat.StatsD/PooledUdpTransport.cs
@@ -86,10 +86,7 @@ namespace JustEat.StatsD
                     while (_socketPool.Count > 0)
                     {
                         var socket = _socketPool.Pop();
-                        if (socket != null)
-                        {
-                            socket.Dispose();
-                        }
+                        socket?.Dispose();
                     }
                 }
                 catch (ObjectDisposedException)
--- a/src/JustEat.StatsD/PooledUdpTransport.cs
+++ b/src/JustEat.StatsD/PooledUdpTransport.cs
@@ -10,10 +10,10 @@ namespace JustEat.StatsD
     /// </summary>
     public sealed class PooledUdpTransport : IStatsDTransport, IDisposable
     {
-        private const int PoolSize = 30;
-
         private readonly SimpleObjectPool<Socket> _socketPool
-            = new SimpleObjectPool<Socket>(PoolSize, pool => UdpTransport.CreateSocket());
+            = new SimpleObjectPool<Socket>(
+                Environment.ProcessorCount,
+                pool => UdpTransport.CreateSocket());
 
         private readonly IPEndPointSource _endpointSource;
         private bool _disposed;
@@ -28,6 +28,22 @@ namespace JustEat.StatsD
         public PooledUdpTransport(IPEndPointSource endPointSource)
         {
             _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
+        }
+
+        /// <inheritdoc />
+        public void Send(string metric)
+        {
+            if (string.IsNullOrWhiteSpace(metric))
+            {
+                return;
+            }
+
+            var bytes = Encoding.UTF8.GetBytes(metric);
+            var endpoint = _endpointSource.GetEndpoint();
+
+            var socket = _socketPool.Pop();
+            socket.SendTo(bytes, endpoint);
+            _socketPool.Push(socket);
         }
 
         /// <summary>
@@ -45,34 +61,6 @@ namespace JustEat.StatsD
             GC.SuppressFinalize(this);
         }
 
-        /// <inheritdoc />
-        public void Send(string metric)
-        {
-            if (string.IsNullOrWhiteSpace(metric))
-            {
-                return;
-            }
-
-            var bytes = Encoding.UTF8.GetBytes(metric);
-            var endpoint = _endpointSource.GetEndpoint();
-
-            var socket = _socketPool.Pop();
-
-            if (socket == null)
-            {
-                return;
-            }
-
-            try
-            {
-                socket.SendTo(bytes, endpoint);
-            }
-            finally
-            {
-                _socketPool.Push(socket);
-            }
-        }
-
         /// <summary>
         /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
         /// </summary>
@@ -84,13 +72,15 @@ namespace JustEat.StatsD
         {
             if (!_disposed)
             {
-                Socket socket = null;
-
                 try
                 {
-                    while ((socket = _socketPool.Pop()) != null)
+                    while (_socketPool.Count > 0)
                     {
-                        socket.Dispose();
+                        var socket = _socketPool.Pop();
+                        if (socket != null)
+                        {
+                            socket.Dispose();
+                        }
                     }
                 }
                 catch (ObjectDisposedException)
--- a/src/JustEat.StatsD/SimpleObjectPool.cs
+++ b/src/JustEat.StatsD/SimpleObjectPool.cs
@@ -22,14 +22,14 @@ namespace JustEat.StatsD
             PrePopulate(initialSize);
         }
 
-        private void PrePopulate(int initialSize)
+        private void PrePopulate(int size)
         {
-            for (var i = 0; i < initialSize; ++i)
+            while (Count < size)
             {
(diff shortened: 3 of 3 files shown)


src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -10,10 +10,10 @@ namespace JustEat.StatsD
     /// </summary>
     public sealed class PooledUdpTransport : IStatsDTransport, IDisposable
     {
-        private const int PoolSize = 30;
+        private const int InitialPoolSize = 2;
```
martincostello: Maybe we could do this (or something similar) instead? Kestrel does something similar to determine the default size of its worker thread pool. Then in theory we have enough pre-allocated for all cores to be busy doing one thing simultaneously.

```csharp
private static int InitialPoolSize = Environment.ProcessorCount * 2;
```

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -84,12 +72,11 @@ private void Dispose(bool disposing)
...
             if (!_disposed)
             {
-                Socket socket = null;
-
                 try
                 {
-                    while ((socket = _socketPool.Pop()) != null)
+                    while (_socketPool.Count > 0)
                     {
+                        var socket = _socketPool.Pop();
```
martincostello: The `null`-check has gone. I haven't read the diff of the pool itself yet, but if it can return null we should check so we don't have the potential for a race-condition where between checking the size and calling `Pop()` the pool has emptied.

src/JustEat.StatsD/SimpleObjectPool.cs
```diff
@@ -1,56 +1,58 @@
...
-        /// <returns>	An object or null if the pool has been exhausted. </returns>
+        public int Count => _pool.Count;
+
+        /// <summary>Retrieves an object from the pool if one is available.
+        /// Expand the pool if need be </summary>
+        /// <returns>An object from the pool. </returns>
         public T Pop()
         {
-            if (!_pool.TryTake(out T result))
+            if (_pool.TryTake(out T result))
```
martincostello: I think _technically_ this is a breaking change if anyone is using this class, as before you could use `null` to mean _is empty_ like when a `StreamReader` returns `null`. Now you can't, and you'd just run in a loop forever creating objects.

AnthonySteele: Yep, that exactly what the test did :(

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -77,7 +77,10 @@ private void Dispose(bool disposing)
                     while (_socketPool.Count > 0)
                     {
                         var socket = _socketPool.Pop();
-                        socket.Dispose();
+                        if (socket != null)
+                        {
+                            socket.Dispose();
```
martincostello: `socket?.Dispose();` for terseness?

AnthonySteele:
The Benchmarks:

### netcoreapp2.0


|            Method |      Mean |     Error |    StdDev | Scaled |  Gen 0 | Allocated |
|------------------ |----------:|----------:|----------:|-------:|-------:|----------:|
|       SendStatUdp | 304.52 us | 5.0794 us | 4.7512 us |   1.00 |      - |    1184 B |
| SendStatPooledUdp |  24.07 us | 0.1363 us | 0.1275 us |   0.08 | 0.2136 |     720 B |

### netcoreapp2.1


|            Method |      Mean |     Error |    StdDev | Scaled |  Gen 0 | Allocated |
|------------------ |----------:|----------:|----------:|-------:|-------:|----------:|
|       SendStatUdp | 297.08 us | 1.0552 us | 0.8239 us |   1.00 |      - |    1184 B |
| SendStatPooledUdp |  24.01 us | 0.2712 us | 0.2117 us |   0.08 | 0.2136 |     720 B |

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -57,20 +42,23 @@ public void Send(string metric)
             var endpoint = _endpointSource.GetEndpoint();
 
             var socket = _socketPool.Pop();
+            socket.SendTo(bytes, endpoint);
```
martincostello: This will leak the socket if `SendTo()` throws, so should be:

```csharp
var socket = _socketPool.Pop();

try
{
    socket.SendTo(bytes, endpoint);
}
catch (Exception)
{
    socket.Dispose();
    throw;
}

_socketPool.Push(socket);
```

martincostello:
The benchmarks look like compared to master it's now twice as slow? Scaled was 0.04 for pooling in #81. I realise that it's still super-fast, but that makes it look like the pooling changes actually slow it down.

martincostello:
Results from 589c3b94667d42d21869b6e0d13ea3aa3ba41039:

``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i7-6500U CPU 2.50GHz (Skylake), 1 CPU, 4 logical and 2 physical cores
.NET Core SDK=2.1.302
  [Host]     : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT


```
|            Method |        Mean |      Error |     StdDev | Scaled |  Gen 0 | Allocated |
|------------------ |------------:|-----------:|-----------:|-------:|-------:|----------:|
|       SendStatUdp | 1,129.36 us | 22.0130 us | 24.4674 us |   1.00 |      - |    1184 B |
| SendStatPooledUdp |    47.47 us |  0.6838 us |  0.6396 us |   0.04 | 0.3052 |     720 B |

Results from [PR 81](https://github.com/justeat/JustEat.StatsD/pull/81#issuecomment-405935145):

``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i7-6500U CPU 2.50GHz (Skylake), 1 CPU, 4 logical and 2 physical cores
.NET Core SDK=2.1.302
  [Host]     : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT


```
|            Method |        Mean |     Error |    StdDev |      Median | Scaled |  Gen 0 | Allocated |
|------------------ |------------:|----------:|----------:|------------:|-------:|-------:|----------:|
|       SendStatUdp | 1,029.29 us | 20.496 us | 54.352 us | 1,007.63 us |   1.00 |      - |    1184 B |
| SendStatPooledUdp |    43.16 us |  1.078 us |  1.801 us |    42.79 us |   0.04 | 0.3052 |     720 B |

Basically, not much in it, but a teeny bit slower.

AnthonySteele:
Acceptable IMHO to have a pool that starts smaller, and grows to cope with demand.

martincostello:
Another thought - whilst the pool is now flexibly sized, it's growth is unbounded. Should we put a ceiling on it?

AnthonySteele:
Is it possible to actually force the pool to be large with a test? The `MultipleMetricsCanBeSentWithoutAnExceptionBeingThrownParallel` test on my machine only makes the pool grow to 8 Sockets. I think that's because `Parallel.For` has detected 8 cores and so is using 8 threads. Can you actually get more threads than that concurrently using the 3 lines of code that rent sockets from the pool?

martincostello:
I can't think of a way to do it, other than somehow doing something that makes `SendTo()` hang and suspend the thread (so it's no longer in the pool), and then continuously start new requests so it grows and grows.

If we're 100% sure there's absolutely no way that could ever do that, then fine, but if it's cheap to put in a safety valve for belt-and-braces that might be worth doing in case there's a scenario we've not thought of and suddenly we starve the machine of sockets.

src/JustEat.StatsD/SimpleObjectPool.cs
AnthonySteele: There is now `Pop` and `PopOrCreate`

martincostello:
I think just put a final set of benchmark results here for the latest commit, then this is good to merge 🎉

src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs
```diff
@@ -42,7 +42,7 @@ public static void MultipleMetricsCanBeSentWithoutAnExceptionBeingThrownParallel
 
             using (var target = new PooledUdpTransport(endPointSource))
             {
-                Parallel.For(0, 10_000, (_) =>
+                Parallel.For(0, 10_000, _ =>
```
slang25: I though you and Martin liked to always include braces. Is this a change of heart?

martincostello: I put the braces there, then Anthony took them away again 🤣

AnthonySteele:
Final benchmarks are close to the same:
````

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i7-4712HQ CPU 2.30GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=2.1.302
  [Host]     : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 64bit RyuJIT
  DefaultJob : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 64bit RyuJIT
````

|            Method |      Mean |     Error |    StdDev | Scaled |  Gen 0 | Allocated |
|------------------ |----------:|----------:|----------:|-------:|-------:|----------:|
|       SendStatUdp | 312.26 us | 5.8638 us | 5.4850 us |   1.00 |      - |    1184 B |
| SendStatPooledUdp |  25.14 us | 0.2954 us | 0.2763 us |   0.08 | 0.2136 |     720 B |

File: src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs (hunk 6 above)
Comment (AnthonySteele) [reply src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs]: I'm not a fan of unnecessary braces.

File: src/JustEat.StatsD/SimpleObjectPool.cs [hunk 7]
```diff
@@ -1,56 +1,67 @@
...
+        private void PrePopulate(int size)
+        {
+            while (Count < size)
             {
-                var instance = constructor(this);
+                var instance = _constructor(this);
                 if (instance == null)
                 {
-                    throw new ArgumentException("constructor produced null object", "constructor");
+                    throw new InvalidOperationException("constructor produced null object");
```
martincostello: Whether it's worth actually doing anything about to revert is up for debate, but based on the standards Microsoft adhere to, but I've just noticed that this is a breaking change.

Issue #85, opened 2018-07-19
[Question] PooledUdpTransport with connected sockets


dv00d00:
Addition of a pulled udp was very important and 20 times the throughput increase is fantastic.

I am wondering though if we could push it a bit further.

What if we will pool connected sockets, like this:

```csharp
        private readonly ConcurrentDictionary<IPEndPoint, Lazy<SimpleObjectPool<Socket>>> _pools =
            new ConcurrentDictionary<IPEndPoint, Lazy<SimpleObjectPool<Socket>>>();

...

        private SimpleObjectPool<Socket> GetPool(IPEndPoint endpoint)
        {
            var pool = _pools.GetOrAdd(
                endpoint,
                ep =>
                {
                    return new Lazy<SimpleObjectPool<Socket>>(
                        () =>
                        {
                            var result = new SimpleObjectPool<Socket>(
                                magic_number,
                                _ =>
                                {
                                    var socket = UdpTransport.CreateSocket();
                                    socket.Connect(ep);
                                    return socket;
                                });

                            return result;
                        }, LazyThreadSafetyMode.ExecutionAndPublication);
                });

            return pool.Value;
        }

// consumption

            var endpoint = _endpointSource.GetEndpoint();
            var pool = GetPool(endpoint);
            var socket = pool.Pop();
            socket.Send(data);

```

I've tried this on my machine:

 |Method |     Toolchain |       Mean |      Error |     StdDev |     Median | Scaled | Allocated |
|----------------- |-------------- |-----------:|-----------:|-----------:|-----------:|-------:|----------:|
| Send | .NET Core 2.1 | 474.578 us | 28.0948 us | 41.1810 us | 464.496 us |   1.00 |     400 B |
| SendWithPool | .NET Core 2.1 |  16.735 us |  0.2070 us |  0.2969 us |  16.696 us |   0.04 |     168 B |
| SendWithMegaPool | .NET Core 2.1 |   6.386 us |  0.1625 us |  0.2330 us |   6.308 us |   0.01 |       0 B |

martincostello:
Interesting - though I'm slightly confused as to what's actually making this faster?

We generally only have one endpoint because our Graphite service is only in one place, so based on he benchmark that's currently in `master` I'm not quite following where the performance gain would have come from with more code being added.

dv00d00:
[...]

PR #86, opened 2018-07-19
Connected UDP Sockets Pool

dv00d00:
Thi pull request modifies PooledUdpTransport class by using a pool of connected udp sockets as discussed in #85 

Benchmarks results:

Before:

|          Method |     Mean |     Error |    StdDev |   Median | Scaled |  Gen 0 | Allocated |
|---------------- |---------:|----------:|----------:|---------:|-------:|-------:|----------:|
| SendWithoutPool | 53.76 us | 1.8980 us | 2.5338 us | 51.65 us |   1.00 | 0.1221 |     440 B |
|    SendWithPool | 14.08 us | 0.0811 us | 0.1189 us | 14.05 us |   0.26 | 0.0610 |     208 B |

After:

|Method |      Mean |     Error |    StdDev | Scaled |  Gen 0 | Allocated |
|--------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
|                 Send | 51.714 us | 0.1015 us | 0.0847 us |   1.00 | 0.1221 |     440 B |
|         SendWithPool |  8.890 us | 0.1292 us | 0.1209 us |   0.17 |      - |      40 B |
| SendWithPoolSwitcher | 11.839 us | 0.0425 us | 0.0398 us |   0.23 | 0.0153 |      67 B |

`SendWithPoolSwitcher` bench tests edge case when endpoint changes every milisecond

### DIFF
--- a/src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs
+++ b/src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs
@@ -1,58 +1,86 @@
 using System;
 using System.Net;
-using System.Threading;
+using System.Net.Sockets;
 using System.Threading.Tasks;
 using JustEat.StatsD.EndpointLookups;
 using Xunit;
 
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
+
 namespace JustEat.StatsD
 {
-    public static class WhenUsingPooledUdpTransport
+    public sealed class UdpListener : IDisposable
     {
-        [Fact]
-        public static void AMetricCanBeSentWithoutAnExceptionBeingThrown()
-        {
-            // Arrange
-            var endPointSource = EndpointLookups.EndpointParser.MakeEndPointSource("127.0.0.1", 8125, null);
+        private readonly Socket _socket;
+        private EndPoint _endPoint;
 
-            using (var target = new PooledUdpTransport(endPointSource))
-            {
-                // Act and Assert
-                target.Send("mycustommetric");
-            }
+        public class State
+        {
+            public const int BufSize = 8 * 1024;
+            public readonly byte[] Buffer = new byte[BufSize];
         }
 
-        [Fact]
-        public static void MultipleMetricsCanBeSentWithoutAnExceptionBeingThrownSerial()
+        public UdpListener(int port)
         {
-            // Arrange
-            var endPointSource = EndpointLookups.EndpointParser.MakeEndPointSource("127.0.0.1", 8125, null);
-
-            using (var target = new PooledUdpTransport(endPointSource))
-            {
-                for (int i = 0; i < 10_000; i++)
-                {
-                    // Act and Assert
-                    target.Send("mycustommetric");
-                }
-            }
+            _socket = new Socket(AddressFamily.InterNetwork, SocketType.Dgram, ProtocolType.Udp);
+            _endPoint = new IPEndPoint(IPAddress.Loopback, port);
+            _socket.SetSocketOption(SocketOptionLevel.IP, SocketOptionName.ReuseAddress, true);
         }
 
-        [Fact]
-        public static void MultipleMetricsCanBeSentWithoutAnExceptionBeingThrownParallel()
+        public void Start()
         {
-            // Arrange
-            var endPointSource = EndpointLookups.EndpointParser.MakeEndPointSource("127.0.0.1", 8125, null);
-
-            using (var target = new PooledUdpTransport(endPointSource))
+            var state = new State();
+            AsyncCallback recv = null;
+            
+            _socket.Bind(_endPoint);
+            _socket.BeginReceiveFrom(state.Buffer, 0, State.BufSize, SocketFlags.None, ref _endPoint, recv = (ar) =>
             {
-                Parallel.For(0, 10_000, _ =>
-                {
-                    target.Send("mycustommetric");
-                });
-            }
+                var so = (State)ar.AsyncState;
+                int _ = _socket.EndReceiveFrom(ar, ref _endPoint);
+                _socket.BeginReceiveFrom(so.Buffer, 0, State.BufSize, SocketFlags.None, ref _endPoint, recv, so);
+            }, state);
         }
 
+        public void Dispose()
+        {
+            _socket.Dispose();
+        }
+    }
+
+    public sealed class Servers : IDisposable
+    {
+        private readonly UdpListener _one;
+        private readonly UdpListener _two;
+
+        public Servers()
+        {
+            _one = new UdpListener(8125);
+            _two = new UdpListener(8126);
+
+            Start();
+        }
+
+        public void Start()
+        {
+            _one.Start();
+            _two.Start();
+        }
+
+        public void Dispose()
+        {
+            _one.Dispose();
+            _two.Dispose();
+        }
+    }
+
+    [CollectionDefinition("REAL_UDP", DisableParallelization = true)]
+    public class DatabaseCollection : ICollectionFixture<Servers>
+    {
+    }
+
+    [Collection("REAL_UDP")]
+    public class WhenUsingPooledUdpTransport 
+    {
         private class MilisecondSwitcher : IPEndPointSource
         {
             private readonly IPEndPointSource _endpointSource1;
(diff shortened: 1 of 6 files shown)


dv00d00:
I was able to reproduce test failures on my mac, created a ticket in corefx (https://github.com/dotnet/corefx/issues/31206)

martincostello:
The failed test could just maybe be a manifestation of differences in OS behaviours a bit like how we have to have this code:

https://github.com/justeat/JustEat.StatsD/blob/554633f110b6e9292a5d69055c171e9a0ebc32da/src/JustEat.StatsD/UdpTransport.cs#L52-L56

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -10,13 +11,71 @@ namespace JustEat.StatsD
...
+            public ConnectedPool(IPEndPoint ipEndPoint)
+            {
+                IpEndPoint = ipEndPoint;
+
+                Pool = new SimpleObjectPool<Socket>(
+                    Environment.ProcessorCount,
+                    pool =>
+                    {
+                        var socket = UdpTransport.CreateSocket();
+                        socket.Connect(ipEndPoint);
```
martincostello: Need to `try-catch` around `Connect()` and dispose in the catch block, as otherwise if it fails we'll leak a socket.

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -10,13 +11,71 @@ namespace JustEat.StatsD
     /// </summary>
     public sealed class PooledUdpTransport : IStatsDTransport, IDisposable
     {
-        private readonly SimpleObjectPool<Socket> _socketPool
-            = new SimpleObjectPool<Socket>(
-                Environment.ProcessorCount,
-                pool => UdpTransport.CreateSocket());
+        private sealed class ConnectedPool : IDisposable
```
martincostello: This private class isn't trivial, so can you make it internal and put it in its own file please?

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -40,66 +99,58 @@ public void Send(string metric)
...
             }
 
-            _socketPool.Push(socket);
+            pool.Pool.Push(socket);
         }
-
-        /// <summary>
-        /// Finalizes an instance of the <see cref="PooledUdpTransport"/> class.
-        /// </summary>
-        ~PooledUdpTransport()
```
martincostello: The `IDisposable` implementation should be retained.

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -40,66 +99,58 @@ public void Send(string metric)
 
             var bytes = Encoding.UTF8.GetBytes(metric);
             var endpoint = _endpointSource.GetEndpoint();
+            var pool = GetPool(endpoint);
+            var socket = pool.Pool.PopOrCreate();
```
martincostello: Is there a better name we can have for this, as `pool.Pool` doesn't scan well?

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -40,66 +99,58 @@ public void Send(string metric)
...
 
-        /// <summary>
-        /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
-        /// </summary>
-        /// <param name="disposing">
-        /// <see langword="true" /> to release both managed and unmanaged resources;
-        /// <see langword="false" /> to release only unmanaged resources.
-        /// </param>
-        private void Dispose(bool disposing)
+        private readonly object _lock = new object();
```
martincostello: Field declarations should live together.

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -40,66 +99,58 @@ public void Send(string metric)
...
-                        throw;
-                    }
+                    _pool.Dispose();
+                    _pool = new ConnectedPool(endPoint);
+                    return _pool;
                 }
-
-                _disposed = true;
             }
+
```
martincostello: Blank line not needed.

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -40,66 +99,58 @@ public void Send(string metric)
...
-                catch (ObjectDisposedException)
+            }
+
+            if (endPoint.Equals(_pool.IpEndPoint))
+            {
+                return _pool;
+            }
+            else
+            {
+                lock (_lock)
```
martincostello: I don't think this lock is a good idea, as it's effectively putting a lock around network calls to `Connect()` in the constructor of the pool.

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -40,66 +99,58 @@ public void Send(string metric)
...
                 {
-                    // If the pool has already been disposed by the finalizer, ignore the exception
-                    if (disposing)
-                    {
-                        throw;
-                    }
+                    _pool.Dispose();
+                    _pool = new ConnectedPool(endPoint);
```
martincostello: If any of the connections in the constructor throw, then subsequent calls will return the disposed-of pool, putting everything in a bad state. Something like an optimistic `Interlocked.CompareExchange()` might be better, as that removes the need for a lock and does not corrupt state if the initialization fails.

Note this doesn't check whether the pool that was created before the thread created its pool is connected to the same endpoint.

```csharp
var oldPool = _pool;
var newPool = new ConnectedPool(endPoint);

try
{
    if (Interlocked.CompareExchange(ref _pool, oldPool, newPool) == oldPool)
    {
        We successfully exchanged the pools
        oldPool.Dispose();
    }
    else
    {
        // Something already replaced the pool
        newPool.Dispose();
    }
}
catch (Exception)
{
    newPool.Dispose();
}
```

src/JustEat.StatsD/PooledUdpTransport.cs
slang25: It's a sealed class, so there would be no point to this dispose pattern.

dv00d00:
I've fixed Unix tests by binding loopback ports to a UDP socket as seen in UdpListeners.cs

src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs
```diff
@@ -1,15 +1,26 @@
...
+        }
+
         [Fact]
-        public static void AMetricCanBeSentWithoutAnExceptionBeingThrown()
+        public void AMetricCanBeSentWithoutAnExceptionBeingThrown()
         {
             // Arrange
-            var endPointSource = EndpointLookups.EndpointParser.MakeEndPointSource("127.0.0.1", 8125, null);
+            var endPointSource = EndpointParser.MakeEndPointSource(
+                UdpListeners.EndpointA.Address.ToString(),
```
martincostello: This makes me wonder whether we should have an overload that accepts an `IPEndPoint`.

src/JustEat.StatsD/ConnectedSocketPool.cs
```diff
@@ -0,0 +1,89 @@
...
+                {
+                    var socket = UdpTransport.CreateSocket();
+                    try
+                    {
+                        socket.Connect(ipEndPoint);
+                        return socket;
+                    }
+                    catch
+                    {
+                        socket?.Dispose();
```
martincostello: `socket` won't ever be `null` here.

src/JustEat.StatsD/ConnectedSocketPool.cs
```diff
@@ -0,0 +1,89 @@
...
+                        throw;
+                    }
+                });
+        }
+
+        public IPEndPoint IpEndPoint { get; }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
martincostello: Can you before & after the perf for with an without having to add these attributes? I'm curious as to what difference they make.

dv00d00: Attribute did not make any difference because method was already inlined by compiler, but this is the difference between inlined

 |Method |     Mean |     Error |    StdDev | Allocated |
|------- |---------:|----------:|----------:|----------:|
|    Act | 79.10 ns | 0.7256 ns | 0.6433 ns |       0 B |

and MethodImplOptions.NoInlining methods

| Method |     Mean |    Error |   StdDev | Allocated |
|------- |---------:|---------:|---------:|----------:|
|    Act | 84.92 ns | 1.681 ns | 2.186 ns |       0 B |

dv00d00: Code for benchmark was:

```csharp

    [MemoryDiagnoser]
    public class InliningBenchmark
    {
        private static readonly ConnectedSocketPool _csp =
            new ConnectedSocketPool(new IPEndPoint(IPAddress.Loopback, 1234));

        [Benchmark]
        public void Act()
        {
            var socket = _csp.PopOrCreate();
            _csp.Push(socket);
        }
    }

```

src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs
martincostello: What do you think @AnthonySteele ?

AnthonySteele: So the fix for the tests on mac and linux was purely in the test setup?

martincostello:
Thank you for your contribution @dv00d00! 🥇

src/JustEat.StatsD.Tests/WhenUsingPooledUdpTransport.cs
martincostello: #92

PR #92, opened 2018-07-30
Add convenience overload of EndpointParser.MakeEndPointSource()

martincostello:
Add a convenience overload for `EndpointParser.MakeEndPointSource()` that accepts an `IPEndpoint` as discussed in #86.

### DIFF
--- a/src/JustEat.StatsD/EndpointLookups/EndpointParser.cs
+++ b/src/JustEat.StatsD/EndpointLookups/EndpointParser.cs
@@ -1,10 +1,20 @@
-﻿using System;
+using System;
 using System.Net;
 
 namespace JustEat.StatsD.EndpointLookups
 {
     public static class EndpointParser
     {
+        public static IPEndPointSource MakeEndPointSource(IPEndPoint endpoint, TimeSpan? endpointCacheDuration)
+        {
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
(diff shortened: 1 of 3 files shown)


src/JustEat.StatsD/EndpointLookups/EndpointParser.cs
```diff
@@ -1,10 +1,20 @@
...
     public static class EndpointParser
     {
+        public static IPEndPointSource MakeEndPointSource(IPEndPoint endpoint, TimeSpan? endpointCacheDuration)
+        {
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+
+            return MakeEndPointSource(endpoint.Address.ToString(), endpoint.Port, endpointCacheDuration);
```
AnthonySteele: If the `endpointCacheDuration` is null, it should probably `return new SimpleIpEndpoint(endpoint);`

martincostello: Actually, on further review, these implementations are backwards. The logic in the other overload should live in this method, and the other method should just call this one. Otherwise we're round-tripping a parse of an `IPAddress` pointlessly.

martincostello: Updated in f165ba94a1a7f0d2c63ed4a61fda6720264adb13.

PR #94, opened 2018-07-31
Version 3.1 beta 01

AnthonySteele:

version metadata so that current changes can be released as a beta for test

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>3.1.0</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
+    <VersionSuffix>beta-01</VersionSuffix>
 
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">$(APPVEYOR_BUILD_NUMBER)</BuildNumber>
     <BuildNumber Condition=" '$(BuildNumber)' == '' ">0</BuildNumber>
--- a/version.props
+++ b/version.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>3.1.0</VersionPrefix>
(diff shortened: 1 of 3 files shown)


AnthonySteele:
So what should `SetAppVeyorBuildVersion.ps1` change to? Is there an example? You don't want a `-something` suffix on every tagged build, as many of them will not be betas. 

I have checked a few other repos (NLog.StructuredLogging.Json, JustSaying) and they do the same as this.

I'm happy for this to be improved some time, but I don't know what to, and want to get code out.

martincostello:
Pseudo code:

```ps
else {
    if ($APPVEYOR_REPO_TAG_NAME.Contains("-")) {
        $versionSuffix = $APPVEYOR_REPO_TAG_NAME.Substring($APPVEYOR_REPO_TAG_NAME.IndexOf("-") +1)
    }
}
```

martincostello:
This doesn't hold you trying it out either, as the AppVeyor build already publishes every build to a pseudo-private NuGet feed you can use internally which you can get the URL of from the NuGet tab in the build's settings over in AppVeyor.

AnthonySteele:
The Pseudocode appears to be correct.

martincostello:
Looks like I missed prefixing the values with `$env:`.

martincostello:
So: `$env:APPVEYOR_REPO_TAG_NAME`

version.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>3.1.0</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
+    <VersionSuffix>beta-01</VersionSuffix>
```
martincostello: Just need to re-blank this now.

AnthonySteele:
The recommendation is to start tags with a `v`, so it should be `v3.1.0-beta01` - it's not consistent, but that is the prefered one.

PR #96, opened 2018-07-31
No auto push

AnthonySteele:

Appveyor config change 

do not automatically push to nuget even on a tag. This is unnecessary, even harmful.
you can do it in the appveyor ui in 2 clicks, and  it's occasional enough and important enough to first verify the package version.

Yes, it has gone wrong today

I am not 100% sure if this is the right syntax. Please comment if you know.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,11 +11,6 @@ build_script:
 artifacts:
 - path: '**\$(APPVEYOR_PROJECT_NAME)*.nupkg'
   name: Nuget
-deploy:
-- provider: NuGet
-  api_key:
-    secure: 6MzbzEs4YdJKS67Gio5gEO8mNKmwfC4UHTCmECZ1KOutI6ndm4vAECazmVNB6an7
-  artifact: /.*nupkg/
 notifications:
 - provider: HipChat
   room: 'Eng :: Open Source'


AnthonySteele:
Does this turn it off, or turn it on when there's _no_ tag:
````
on:	
    APPVEYOR_REPO_TAG: false
```

martincostello:
I assume that means “On, but not for tags”.

martincostello:
I have an idea to test this out:
  1. We comment out the whole NuGet publish section from appveyor.yml.
  1. We add an extra line to [`version.props`](https://github.com/justeat/JustEat.StatsD/blob/master/version.props) to set the version suffix appropriately automatically for tagged builds.
  1. Merge.
  1. Add a tag.
  1. See what gets created.

If successful, we just do a PR to restore the publish configuration, and then we can safely release from tags.

I'm happy to do the work to do the above if you think it's a good idea.

AnthonySteele:
Or we could ask them?

martincostello:
You could invent a branch that doesn't exist? https://www.appveyor.com/docs/deployment/#deploy-on-tag-github-and-gitlab-only

AnthonySteele:
https://help.appveyor.com/discussions/questions/23546-nuget-manual-publish-syntax

AnthonySteele:
As per help.apveyor.com thread, support person was finally forced to give a simple answer to a simple question: manual deploy is configured separately, via the "environments" gui, and you can disable autopush by deleting the whole "deploy" section from `appveyor.yml`

PR #97, opened 2018-07-31
Beta 01 suffix

AnthonySteele:

Prepare for v `3.1.0-beta01`

Getting tagging working right is out of scope and can be done later and separately. but only after automatic deployment to nuget has been turned off.

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>3.1.0</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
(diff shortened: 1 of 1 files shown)


martincostello:
Change in #98 should fix the tag not being used to version the `.nupkg` file.

AnthonySteele:
So I just want to get a beta out. This PR allows it now, while getting tags to work is still ongoing.

martincostello:
I'm fairly confident now that #98 is merged, this would work by just creating a tag without merging this or #96, but you can merge it if you want to.

PR #101, opened 2018-08-15
Update .NET Core SDK

martincostello:
Update to .NET Core SDK `2.1.400`.


martincostello:
Looks like I need to apply a fix to `Build.ps1` to handle a difference in SDK behaviour over time.

Issue #103, opened 2018-09-05
Unclear meaning of DateTime for Gauges


dv00d00:
There are two methods in `StatsDMessageFormatter` class which are using DateTime parameter

``` csharp

public string Gauge(double magnitude, string statBucket, DateTime timestamp)
{
    var stat = string.Format(InvariantCulture, "{0}{1}:{2}|g|@{3}", _prefix, statBucket, magnitude, timestamp.AsUnixTime());
    return Format(DefaultSampleRate, stat);
}

public string Gauge(long magnitude, string statBucket, DateTime timestamp)
{
    var stat = string.Format(InvariantCulture, "{0}{1}:{2}|g|@{3}", _prefix, statBucket, magnitude, timestamp.AsUnixTime());
    return Format(DefaultSampleRate, stat);
}

```

I was not able to find any reference to Unix timestamp in statsd docs

https://github.com/b/statsd_spec#gauges
https://statsd.readthedocs.io/en/v3.3/types.html#gauges
https://github.com/etsy/statsd/blob/master/docs/metric_types.md#gauges

Also this paragraph 

```
The StatsClient.gauge() method also support the rate parameter to sample data back to the statsd server, but use it with care, especially with gauges that may not be updated very often.
```

Am I missing something?

slang25:
Agreed, it looks weird. Can't see any real justification in the (very old) commit that added it: https://github.com/justeat/JustEat.StatsD/commit/63c1a4517ed133cc479465d2d56f097703053ba7

martincostello:
I'd concur that we could drop it as part of `4.0.0`.

PR #104, opened 2018-09-05
Zero allocating StatsDPublisher

dv00d00:
I've added a proof of concept implementation of buffer based statsd publisher.

Key points:

- A concept of statsd message struct
- Do not attempt formatting before sampling
- Format directly into utf8
- v2 transport interface uses ArraySegment<byte> which was in .net since 4.5
- zero allocations in .net core 2.1 with pooled udp transport

`Utf8FormatterBench`

|Method |     Mean |     Error |    StdDev | Scaled |  Gen 0 | Allocated |
|------------ |---------:|----------:|----------:|-------:|-------:|----------:|
| StringBased | 609.8 ns | 2.1574 ns | 0.1219 ns |   1.00 | 0.2203 |     696 B |
| BufferBased | 279.9 ns | 0.1432 ns | 0.0081 ns |   0.46 |      - |       0 B |

`StatSendingBenchmark`

|Method |       Mean |     Error |    StdDev |  Gen 0 | Allocated |
|------------------- |-----------:|----------:|----------:|-------:|----------:|
|             RunUdp | 288.132 us | 6.1692 us | 0.3486 us | 0.4883 |    2944 B |
| RunUdpWithSampling |  30.062 us | 2.3620 us | 0.1335 us | 0.1221 |     467 B |
|              RunIp | 264.734 us | 9.0817 us | 0.5131 us | 0.4883 |    2944 B |
|  RunIpWithSampling |  29.965 us | 2.8734 us | 0.1624 us | 0.1221 |     468 B |
|              RunV2 |  62.660 us | 1.9967 us | 0.1128 us |      - |       0 B |
|  RunV2WithSampling |   6.438 us | 0.3993 us | 0.0226 us |      - |       0 B |


dv00d00:
Open questions:

- The fate of TimeStamp in gauges, see #103 
- Max size of the encoded message. In theory, UDP datagrams could be as large as 8kb, in practice its usually 512 bytes max. Should we add an invariant on it? 
- Gauge does support sampling in statsd, should we expose it?

src/Benchmark/StatSendingBenchmark.cs
```diff
@@ -2,13 +2,16 @@
 using BenchmarkDotNet.Attributes;
 using JustEat.StatsD;
 using JustEat.StatsD.EndpointLookups;
+using JustEat.StatsD.V2;
```
slang25: Perhaps the new publisher, transport types and namespaces should have a name other than V2, naming stuff is hard though.

Something that indicates it offers better perf / zero allocations.

slang25:
This PR is fantastic work @dv00d00, zero allocations and it's not a breaking change! 💯

src/Benchmark/StatSendingBenchmark.cs
martincostello: +1, V2 is confusing given that the library is at version 3.

Directory.Build.props
```diff
@@ -24,5 +24,6 @@
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>$(PackageProjectUrl).git</RepositoryUrl>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <LangVersion>7.3</LangVersion>
```
martincostello: This is redundant, we already use the latest version here: https://github.com/justeat/JustEat.StatsD/blob/a3ee3711594b048e94be4068fba54dffd6c63094/Directory.Build.props#L14

src/JustEat.StatsD.Tests/Utf8FormatterTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            Check(message, "prefix.bucket:128|ms");
+        }
+
+        [Fact]
+        public static void TimingSampled()
+        {
+            var message = StatsDMessage.Timing(128, "bucket");
+            Check(message, 0.5, "prefix.bucket:128|ms|@0.5");
+        }
+
```
martincostello: Extra new line.

src/JustEat.StatsD.Tests/Utf8FormatterTests.cs
```diff
@@ -0,0 +1,76 @@
...
+        private static void Check(StatsDMessage message,  string expected)
+        {
+            Check(message, 1, expected);
+        }
+
+        private static void Check(StatsDMessage message, double sampleRate, string expected)
+        {
+            Formatter.TryFormat(message, sampleRate, Buffer, out int written).ShouldBe(true);
+            var result = Encoding.UTF8.GetString(Buffer.AsSpan().Slice(0, written));
```
martincostello: I _think_ there's an `AsSpan()` overload that lets you do this in one:

```csharp
var result = Encoding.UTF8.GetString(Buffer.AsSpan(0, written));
```

src/JustEat.StatsD/JustEat.StatsD.csproj
```diff
@@ -5,14 +5,19 @@
     <PackageId>JustEat.StatsD</PackageId>
     <RootNamespace>JustEat.StatsD</RootNamespace>
     <Summary>A .NET library for publishing metrics to statsd.</Summary>
-    <TargetFrameworks>net451;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net451;netstandard2.0;netcoreapp2.1</TargetFrameworks>
```
martincostello: I thought we were going to avoid adding extra TFMs?

src/JustEat.StatsD/Properties/AssemblyInfo.cs
```diff
@@ -1,4 +1,8 @@
-﻿using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 
 [assembly: ComVisible(false)]
 [assembly: Guid("8f4ff09e-4130-4872-a50f-b290e9ccb04b")]
+
+[assembly:InternalsVisibleTo("JustEat.StatsD.Tests")]
```
martincostello: Spaces after colons.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            Tail = source;
+            Written = 0;
+        }
+
+        public Span<byte> Tail;
+        public int Written;
+    }
+
+    internal static class BufferImpl
```
martincostello: One type per file please.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteBytes(this ref Buffer src, Span<byte> bytes)
```
martincostello: Would `destination` be a better name instead of `bytes`?

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+#if NETCOREAPP2_1
+            try
+            {
+                var written = Encoding.UTF8.GetBytes(str, src.Tail);
+                src.Tail = src.Tail.Slice(written);
+                src.Written += written;
+                return true;
+            }
+            catch
```
martincostello: We shouldn't blindly catch any exception.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+            src.Written += bytes.Length;
+            return true;
+        }
+
+        public static bool TryWriteUtf8String(this ref Buffer src, string str)
+        {
+#if NETCOREAPP2_1
+            try
+            {
+                var written = Encoding.UTF8.GetBytes(str, src.Tail);
```
martincostello: Would storing `Encoding.UTF8` in a static field improve the perf slightly by removing the need to keep accessing it through the getter?

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+                return false;
+            }
+#else
+            var bucketBytes = Encoding.UTF8.GetBytes(str);
+            return src.TryWriteBytes(bucketBytes);
+#endif
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteBytes(this ref Buffer src, byte chr)
```
martincostello: `ch` is more idiomatic.

src/JustEat.StatsD/V2/Buffer.cs
martincostello: `BufferExtensions` would be a better name given that this class entirely contains extension methods.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+        public Span<byte> Tail;
+        public int Written;
+    }
+
+    internal static class BufferImpl
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteBytes(this ref Buffer src, Span<byte> bytes)
+        {
+            if (bytes.Length > src.Tail.Length) return false;
```
martincostello: Don't elide braces please:

```csharp
if (...)
{
   ...
}
```

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+            var bucketBytes = Encoding.UTF8.GetBytes(str);
+            return src.TryWriteBytes(bucketBytes);
+#endif
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteBytes(this ref Buffer src, byte chr)
+        {
+            const int length = 1;
+            if (src.Tail.Length < length) return false;
```
martincostello: As above.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+#else
+            var bucketBytes = Encoding.UTF8.GetBytes(str);
+            return src.TryWriteBytes(bucketBytes);
+#endif
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteBytes(this ref Buffer src, byte chr)
+        {
+            const int length = 1;
```
martincostello: `OneByte`?

src/JustEat.StatsD/V2/Buffer.cs
martincostello: `TryWriteByte()` (not plural).

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            src.Tail[0] = chr;
+            src.Tail = src.Tail.Slice(length);
+            src.Written += length;
+
+            return true;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteBytes(this ref Buffer src, byte chr1, byte chr2)
```
martincostello: `ch1`, `ch2`, or, `first` and `second`.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            const int length = 2;
```
martincostello: `TwoBytes`?

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+            if (src.Tail.Length < length) return false;
```
martincostello: As above.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+            src.Tail[0] = chr1;
+            src.Tail[1] = chr2;
+            src.Tail = src.Tail.Slice(length);
+            src.Written += length;
+
+            return true;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteBytes(this ref Buffer src, byte chr1, byte chr2, byte chr3)
```
martincostello: Same comments as above method regarding naming and braces.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+            src.Tail[1] = chr2;
+            src.Tail[2] = chr3;
+            src.Tail = src.Tail.Slice(length);
+            src.Written += length;
+
+            return true;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteLong(this ref Buffer src, long val)
```
martincostello: `TryWriteInt64()`

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            if (!Utf8Formatter.TryFormat(val, src.Tail, out var consumed)) return false;
```
martincostello: Braces.

src/JustEat.StatsD/V2/Buffer.cs
```diff
@@ -0,0 +1,115 @@
...
+            src.Tail = src.Tail.Slice(consumed);
+            src.Written += consumed;
+
+            return true;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryWriteDouble(this ref Buffer src, double val)
+        {
+            if (!Utf8Formatter.TryFormat((decimal) val, src.Tail, out var consumed)) return false;
```
martincostello: Braces.

src/JustEat.StatsD/V2/PooledUdpTransportV2.cs
```diff
@@ -0,0 +1,70 @@
+using System;
+using System.Net;
+using System.Net.Sockets;
+using System.Threading;
+using JustEat.StatsD.EndpointLookups;
+
+namespace JustEat.StatsD.V2
+{
+    public sealed class PooledUdpTransportV2 : IDisposable, IStatsDTransportV2
```
martincostello: Why not make `PooledUdpTransport` just implement `IStatsDTransport` _and_ `IStatsDTransportV2`? Then we don't have two types with almost the same implementation.

src/JustEat.StatsD/V2/StatsDMessage.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            return new StatsDMessage(statBucket, magnitude, Kind.Counter);
+        }
+
+        public static StatsDMessage Gauge(double magnitude, string statBucket)
+        {
+            return new StatsDMessage(statBucket, magnitude, Kind.Gauge);
+        }
+
+        internal enum Kind { Counter, Timing, Gauge }
```
martincostello: Own file and type, not nested.

martincostello: Declare one value per line.

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
```diff
@@ -0,0 +1,170 @@
+using System;
+using System.Threading;
+using JustEat.StatsD.EndpointLookups;
+
+namespace JustEat.StatsD.V2
+{
+    public sealed class StatsDPublisherV2 : IStatsDPublisher
```
martincostello: I feel that it would be better to have the original publisher's implementation details be able to handle `IStatsDPublisher` and `IStatsDPublisherV2` so we don't need all these "V2" alternate versions which double the amount of code to maintain.

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
```diff
@@ -0,0 +1,170 @@
...
+    {
+        private const int DefaultSampleRate = 1;
+
+        private static int _bufferSize = 512; // safe max size of udp packet
+
+        [ThreadStatic]
+        private static byte[] _buffer;
+        private static byte[] Buffer() => _buffer ?? (_buffer = new byte[_bufferSize]);
```
martincostello: Could just make this a property?

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        [ThreadStatic]
+        private static Random _random;
+        private static Random Random() => _random ?? (_random = new Random());
```
martincostello: Could just make this a property?

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
```diff
@@ -0,0 +1,170 @@
...
+            }
+
+            var endpointSource = EndpointParser.MakeEndPointSource(
+                configuration.Host, configuration.Port, configuration.DnsLookupInterval);
+
+            _transport = new PooledUdpTransportV2(endpointSource);
+            _formatter = new StatsDUtf8Formatter(configuration.Prefix);
+        }
+
+        public StatsDPublisherV2(StatsDConfiguration configuration, IStatsDTransportV2 transport)
```
martincostello: For example, if we ported this into the existing publisher, you could do something like:

```
public StatsDPublisher(StatsDConfiguration configuration)
{
    // Do "v2" stuff
}

public StatsDPublisher(StatsDConfiguration configuration, IStatsDTransport transport)
{
    // Validation goes here...

    if (transport is IStatsDTransportV2 transport2)
    {
        // Do "v2" stuff with transport2
    }
    else
    {
        // Do "v1" stuff
    }
}

public StatsDPublisher(StatsDConfiguration configuration, IStatsDTransportV2 transport)
{
    // Do "v2" stuff
}
```

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (sampleRate >= 1 || sampleRate >= Random().NextDouble())
+            {
+                var msg = StatsDMessage.Counter(value, bucket);
+                SendMessage(sampleRate, msg);
+            }
+        }
+
+        public void Increment(long value, double sampleRate, params string[] buckets)
+        {
+            foreach (var bucket in buckets)
```
martincostello: No null check on `buckets`.

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        public void Decrement(long value, double sampleRate, string bucket)
+        {
+            Increment(value > 0 ? -value : value, sampleRate, bucket);
+        }
+
+        public void Decrement(long value, double sampleRate, params string[] buckets)
+        {
+            foreach (var bucket in buckets)
```
martincostello: No null check on buckets.

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            if (_formatter.TryFormat(msg, sampleRate, buffer, out int written))
+            {
+                _transport.Send(new ArraySegment<byte>(buffer, 0, written));
+            }
+            else
+            {
+                var newSize = _formatter.GetBufferSize(msg);
+                var size = Interlocked.CompareExchange(ref _bufferSize, buffer.Length, newSize);
+                _buffer = new byte[size];
```
martincostello: It's a bit weird here that we change `_buffer` and then call it through `Buffer()`, rather than just use it.

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
```diff
@@ -0,0 +1,170 @@
...
+
+                if (_formatter.TryFormat(msg, sampleRate, Buffer(), out written))
+                {
+                    _transport.Send(new ArraySegment<byte>(buffer, 0, written));
+                }
+                else
+                {
+                    throw new Exception("TODO");
```
martincostello: `FIXME`

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace JustEat.StatsD.V2
+{
+    internal class StatsDUtf8Formatter
+    {
+        private readonly byte[] _utf8Prefix;
+
+        public StatsDUtf8Formatter(string prefix = "")
+        {
+            _utf8Prefix = string.IsNullOrWhiteSpace(prefix) ? new byte[0] : Encoding.UTF8.GetBytes(prefix + ".");
```
martincostello: Use `Array.Empty<byte>()`?

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
+using System;
+using System.Text;
+
+namespace JustEat.StatsD.V2
+{
+    internal class StatsDUtf8Formatter
+    {
+        private readonly byte[] _utf8Prefix;
+
+        public StatsDUtf8Formatter(string prefix = "")
```
martincostello: Do we really need this optional constructor parameter?

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            _utf8Prefix = string.IsNullOrWhiteSpace(prefix) ? new byte[0] : Encoding.UTF8.GetBytes(prefix + ".");
+        }
+
+        public int GetBufferSize(in StatsDMessage msg)
+        {
+            const int maxSerializedDoubleSymbols = 20;
```
martincostello: Constants should start with a capital letter.

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+            const int maxUtf8BytesPerChar = 4;
+            const int colonBytes = 1;
+
+            const int maxMessageKindSuffixSize = 3;
+            const int maxSamplingSuffixSize = 2;
+
+            return _utf8Prefix.Length + msg.StatBucket.Length * maxUtf8BytesPerChar
```
martincostello: Add parentheses to make the maths on this line more explicit to the reader.

```csharp
(_utf8Prefix.Length + msg.StatBucket.Length) * maxUtf8BytesPerChar
// or
_utf8Prefix.Length + (msg.StatBucket.Length * maxUtf8BytesPerChar)
```

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+              + colonBytes
+              + maxSerializedDoubleSymbols
+              + maxMessageKindSuffixSize
+              + maxSamplingSuffixSize
+              + maxSerializedDoubleSymbols;
+        }
+
+        // prefix + msg.Bucket + ":" + msg.Value + (<oneOf> "|ms", "|c", "|g") + {<optional> "|@" + sampleRate }
```
martincostello: Put this comment _in_ the method.

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+        public bool TryFormat(in StatsDMessage msg, double sampleRate, Span<byte> destination, out int written)
+        {
+            written = 0;
+            var buffer = new Buffer(destination);
+
+            if (!buffer.TryWriteBytes(_utf8Prefix)) return false;
```
martincostello: Braces please.

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+            if (!buffer.TryWriteUtf8String(msg.StatBucket)) return false;
+            if (!buffer.TryWriteBytes((byte)':')) return false;
```
martincostello: Could make `(byte)';'` a constant, unless the compiler does that already.

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+                    else
+                    {
+                        if (!buffer.TryWriteDouble(msg.Magnitude)) return false;;
+                    }
+
+                    if (!buffer.TryWriteBytes((byte)'|', (byte)'g')) return false;
+                    break;
+                }
+                default:
+                    throw new ArgumentOutOfRangeException();
```
martincostello: `NotImplementedException`?

martincostello: Add a message so that if it happens we know _what_ happened.

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+            if (!buffer.TryWriteBytes((byte)':')) return false;
+
+            var magnitudeIntegral = (long) msg.Magnitude;
+
+            switch (msg.MessageKind)
+            {
+                case StatsDMessage.Kind.Counter:
+                {
+                    if (!buffer.TryWriteLong(magnitudeIntegral)) return false;
+                    if (!buffer.TryWriteBytes((byte)'|', (byte)'c')) return false;
```
martincostello: Constants?

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+                    break;
+                }
+                case StatsDMessage.Kind.Timing:
+                {
+                    if (!buffer.TryWriteLong(magnitudeIntegral)) return false;
+                    if (!buffer.TryWriteBytes((byte)'|', (byte)'m', (byte)'s')) return false;
```
martincostello: Constants?

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+                    if (msg.Magnitude == (double)magnitudeIntegral)
+                    {
+                        if (!buffer.TryWriteLong(magnitudeIntegral)) return false;
+                    }
+                    else
+                    {
+                        if (!buffer.TryWriteDouble(msg.Magnitude)) return false;;
+                    }
+
+                    if (!buffer.TryWriteBytes((byte)'|', (byte)'g')) return false;
```
martincostello: Constants?

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+                    break;
+                }
+                default:
+                    throw new ArgumentOutOfRangeException();
+            }
+
+            if (sampleRate < 1.0 && sampleRate > 0.0)
+            {
+                if (!buffer.TryWriteBytes((byte)'|', (byte)'@')) return false;
```
martincostello: Constants?

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,87 @@
...
+                if (!buffer.TryWriteDouble(sampleRate)) return false;
+            }
+
+            written = buffer.Written;
+            return true;
+        }
+
```
martincostello: Extra new line.

src/JustEat.StatsD/V2/UdpTransportV2.cs
```diff
@@ -0,0 +1,50 @@
...
+    public sealed class UdpTransportV2 : IStatsDTransportV2
+    {
+        private readonly IPEndPointSource _endpointSource;
+
+        public UdpTransportV2(IPEndPointSource endPointSource)
+        {
+            _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
+        }
+
+        public void Send(ArraySegment<byte> metric)
```
martincostello: As per previous comments, we should be able to make the existing types implement both ways.

src/JustEat.StatsD/JustEat.StatsD.csproj
slang25: To fully take advantage of `Span<T>` we'll want to use `netcoreapp2.1`, or wait until `netstandard2.1` (which `netcoreapp2.1` won't implement btw! only `netcoreapp3.0` and up). 

If we wait until `netstandard2.1` we will want to multi-target to keep support for current clients, so we may as well embrace `netcoreapp2.1` today is my thought.

martincostello: I meant more non-standard TFMs (i.e. the app ones), rather than any TFMs at all.

Is there some guidance somewhere on advantages vs. disadvantages to targeting the more specific implementations?

slang25: There will be guidance from Microsoft soon apparently, still waiting.

There are pros and cons, the bottom line is if you want to use the new APIs today, you have to target the platforms directly to use them. netstandard will always trail behind. For this sort of library, it feels appropriate to add this target.

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
dv00d00: needs 4.6 at least, we are pointing to 4.5.1

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
dv00d00: It does, at least saw this trick in various span based serializer

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
dv00d00: would add 50 lines of code, would definitely make it harder to read. I dig the default coding guidelines. But here `if (false) return` is more like F# `do!` (monadic bind) in a sense that it terminates computation,

dv00d00: I'll update if you insist though

dv00d00:
I am going to address V2 naming and structure in upcoming commits

src/JustEat.StatsD/V2/StatsDUtf8Formatter.cs
martincostello: Could use it for the newer TFMs if we're going to go down the `#if FRAMEWORK` and trying to reduce allocations though...

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
dv00d00: could only be applied to fields

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
dv00d00: could only be applied to fields

src/JustEat.StatsD/V2/StatsDPublisherV2.cs
martincostello: Should be able to do this?

```csharp
private static Random Random => _random ?? (_random = new Random());
```

dv00d00:
Ok, I think I am done with refactoring and addressing issues, would appreciate another review

src/JustEat.StatsD.Tests/WhenTheTransportThrows.cs
```diff
@@ -1,66 +1,81 @@
 using System;
 using System.Net.Sockets;
+using JustEat.StatsD.Buffered;
 using Shouldly;
 using Xunit;
 
+#pragma warning disable xUnit1026 // Theory methods should use all of their parameters disabled to render test case name
```
martincostello: Can you provide a screenshot of the before and after with this warning disabled?

src/JustEat.StatsD.Tests/WhenTheTransportThrows.cs
```diff
@@ -1,66 +1,81 @@
...
             throw new SocketException(42);
         }
+
+        public void Send(ArraySegment<byte> metric)
+        {
+            throw new SocketException(42);
+        }
     }
+    
+    public delegate IStatsDPublisher Factory(StatsDConfiguration config);
```
martincostello: Could you not just use `Func<StatsDConfiguration, IStatsDPublisher>`?

src/JustEat.StatsD/Buffered/BufferBasedStatsDPublisher.cs
```diff
@@ -0,0 +1,184 @@
...
+
+        public void Increment(long value, double sampleRate, string bucket)
+        {
+            var msg = StatsDMessage.Counter(value, bucket);
+            SendMessage(sampleRate, msg);
+        }
+
+        public void Increment(long value, double sampleRate, params string[] buckets)
+        {
+            if (buckets == null)
```
martincostello: Could also shortcut if `buckets.Length == 0` to remove the need to allocate the enumerator.

src/JustEat.StatsD/Buffered/BufferBasedStatsDPublisher.cs
```diff
@@ -0,0 +1,184 @@
...
+
+                    if (_formatter.TryFormat(msg, sampleRate, _buffer, out written))
+                    {
+                        _transport.Send(new ArraySegment<byte>(_buffer, 0, written));
+                    }
+                    else
+                    {
+                        // so we was not able to write to resized buffer
+                        // that means there is a bug in formatter
+                        throw new Exception("Utf8 Formatting Error. This is a bug." +
```
martincostello: I'm not sure about throwing vague exceptions to users going "oh, there's a bug".

image

src/JustEat.StatsD/Buffered/BufferExtensions.cs
```diff
@@ -0,0 +1,127 @@
+using System;
+using System.Buffers.Text;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace JustEat.StatsD.Buffered
+{
+    internal static class BufferExtensions
```
martincostello: Suggest adding `[EditorBrowsable(EditorBrowsableState.Never)]` to the class to hide it (but not the methods) from Intellisense.

src/JustEat.StatsD/Buffered/StatsDMessageKind.cs
```diff
@@ -0,0 +1,9 @@
+namespace JustEat.StatsD.Buffered
+{
+    internal enum StatsDMessageKind
+    {
+        Counter,
+        Timing,
+        Gauge
```
martincostello: Should add a `,` to the end of the line so that adding a new member in the future is a 1 line diff, not 2.

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,105 @@
...
+
+namespace JustEat.StatsD.Buffered
+{
+    internal class StatsDUtf8Formatter
+    {
+        private readonly byte[] _utf8Prefix;
+
+        public StatsDUtf8Formatter(string prefix)
+        {
+            _utf8Prefix = string.IsNullOrWhiteSpace(prefix) ? new byte[0] : Encoding.UTF8.GetBytes(prefix + ".");
```
martincostello: What about:

```csharp
_utf8Prefix =
    string.IsNullOrWhiteSpace(prefix) ?
#if NET451
    new byte[0] :
#else
    Array.Empty<byte>() :
#endif
    Encoding.UTF8.GetBytes(prefix + ".");
```

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,105 @@
...
+        public int GetMaxBufferSize(in StatsDMessage msg)
+        {
+            const int MaxSerializedDoubleSymbols = 32;
+            const int ColonBytes = 1;
+
+            const int MaxMessageKindSuffixSize = 3;
+            const int MaxSamplingSuffixSize = 2;
+
+            return _utf8Prefix.Length
+                    + Encoding.UTF8.GetByteCount(msg.StatBucket)
```
martincostello: Do the same caching of the encoding into a static field?

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
```diff
@@ -0,0 +1,105 @@
...
+                    bool isMagnitudeIntegral = msg.Magnitude == integralMagnitude;
+
+                    var successSoFar = isMagnitudeIntegral ?
+                        buffer.TryWriteInt64(integralMagnitude) :
+                        buffer.TryWriteDouble(msg.Magnitude);
+
+                    return successSoFar && buffer.TryWriteBytes((byte) '|', (byte) 'g');
+                }
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(msg.MessageKind), $"Unknown StatsD message kind encountered {msg.MessageKind}");
```
martincostello: Technically the argument name is `msg`, not `MessageKind`.

src/JustEat.StatsD/Buffered/IStatsDBufferedTransport.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace JustEat.StatsD.Buffered
+{
+    internal interface IStatsDBufferedTransport
```
martincostello: Did you mean to make this `internal`? Otherwise the transports only support it when using the concrete type.

src/JustEat.StatsD/IpTransport.cs
```diff
@@ -30,6 +31,20 @@ public void Send(string metric)
             }
         }
 
+        public void Send(ArraySegment<byte> metric)
+        {
+            if (metric.Count == 0)
```
martincostello: Can `ArraySegment<>` be `null`, or is it a `struct`?

src/JustEat.StatsD/JustEat.StatsD.csproj
```diff
@@ -5,14 +5,19 @@
...
     <PackageReference Include="Microsoft.NETCore.Platforms" Version="1.0.2" />
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
+    <PackageReference Include="System.Memory" Version="4.5.1" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="System.Memory" Version="4.5.1" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.0.0" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.1' ">
```
martincostello: Could change this to `Condition=" '$(TargetFramework)' != 'net451' "` and then remove the need to reference it twice for both `netstandard2.0` and `netcoreapp2.1`.

src/JustEat.StatsD/PooledUdpTransport.cs
```diff
@@ -51,6 +53,29 @@ public void Send(string metric)
             pool.Push(socket);
         }
 
+        public void Send(ArraySegment<byte> metric)
+        {
+            if (metric.Count == 0)
```
martincostello: Same question regarding if it can be `null`.

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -8,139 +9,136 @@ namespace JustEat.StatsD
     /// </summary>
     public class StatsDPublisher : IStatsDPublisher
     {
-        private readonly StatsDMessageFormatter _formatter;
-        private readonly IStatsDTransport _transport;
-        private readonly Func<Exception, bool> _onError;
+        private readonly IStatsDPublisher _publisher;
 
-        public StatsDPublisher(StatsDConfiguration configuration, IStatsDTransport transport)
+        public StatsDPublisher(StatsDConfiguration configuration, IStatsDTransport transport, bool preferBufferedTransport = false)
```
martincostello: This is a binary breaking change - is that intentional?

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -8,139 +9,136 @@ namespace JustEat.StatsD
...
+                    _publisher = new BufferBasedStatsDPublisher(configuration, transportV2);
+                    break;
+                default:
+                    _publisher = new StringBasedStatsDPublisher(configuration, transport);
+                    break;
+            }
         }
 
-        public StatsDPublisher(StatsDConfiguration configuration)
+        public StatsDPublisher(StatsDConfiguration configuration, bool preferBufferedTransport = false)
```
martincostello: As above.

src/JustEat.StatsD/UdpTransport.cs
```diff
@@ -44,6 +45,21 @@ public void Send(string metric)
             }
         }
 
+        public void Send(ArraySegment<byte> metric)
```
martincostello: Null check?

src/JustEat.StatsD/Buffered/BufferExtensions.cs
slang25: Nice tip, never thought of that. I guess it's more important for public types in libraries like this, but still has merit here.

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
slang25: Agreed, and I think just `new ArgumentOutOfRangeException(nameof(msg))` would be more than fine in this instance as it's an internal enum, so this will only get hit if someone adds a new case to this library and doesn't update this case statement.

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
dv00d00: It is not a hot path anyway, created once per app lifetime

src/JustEat.StatsD/UdpTransport.cs
dv00d00: ArraySegmentis a struct and could be constructed with the default constructor, resulting in the null array. I will add a null check to the Array field.

src/JustEat.StatsD/JustEat.StatsD.csproj
martincostello: Looks like the guidance is “_consider_”: https://twitter.com/ben_a_adams/status/1039989566021820418?s=21

src/JustEat.StatsD/PooledUdpTransport.cs
dv00d00: Array which it holds can (if you construct ArraySegment with default constructor) or just use `default(ArraySegment<byte>)`. Will add checks

src/JustEat.StatsD/IpTransport.cs
dv00d00: but array needs to be checked

src/JustEat.StatsD/Buffered/IStatsDBufferedTransport.cs
dv00d00: Making it public

dv00d00:
There is currently no way to specify a constructor on StasDPublisher which would accept IBufferedStatsDTransport, it needs to have the separate signature, otherwise, compiler complains over Ambiguous invocation.

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
martincostello: If it’s consistently slower across several runs, should undo the other one I suggested then.

src/JustEat.StatsD/Buffered/IStatsDBufferedTransport.cs
martincostello: If this is going to be public then I suggest not putting it in the `Buffered` namespace for discoversbility, as otherwise it’s the only public type in that whole namespace.

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -34,7 +42,12 @@ public StatsDPublisher(StatsDConfiguration configuration, IStatsDTransport trans
             }
         }
 
-        public StatsDPublisher(StatsDConfiguration configuration, bool preferBufferedTransport = false)
+        public StatsDPublisher(StatsDConfiguration configuration)
+            : this(configuration, false)
+        {
+        }
+
+        public StatsDPublisher(StatsDConfiguration configuration, bool preferBufferedTransport)
```
martincostello: What's the reasoning behind making "better performance" opt-in? I think it would be better to add a constructor to accept the new interface, and have the existing two try use the better implementation, if possible, by default.

src/JustEat.StatsD/Buffered/IStatsDBufferedTransport.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace JustEat.StatsD.Buffered
```
martincostello: I think this would be better in `JustEat.StatsD` rather than `JustEat.StatsD.Buffered` to make it more discoverable.

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
martincostello: OK, I just figured maybe it was worth doing if the PR is trying to reduce allocations.

src/JustEat.StatsD/StatsDPublisher.cs
dv00d00: As I have added new transport interface implementation to the existing transports, adding a constructor with signature accepting transport would result in a compilation error for the consumer.

I will make buffered implementation a default though.

martincostello: I'm not sure whether it's a good idea or not, but if you swapped the parameter order for the new one, it wouldn't be ambiguous to the compiler?

dv00d00: The compiler will accept it without issues. 😸 
Looks hacky though. 

Other options I can think of: 
- Static factory methods on StatsDPublisher
- Dedicated factory class
- Dedicated transports per interface
- Expose new public implementation of BufferdPublisher, String publisher will retain the old name

martincostello:
So I think just two things left on this for me:
  1. Is there a way we can make the constructor for the publisher accept the new interface without creating ambiguity to the compiler? Maybe we could add a new factory type that could simplify the transport creation and hide the buffering/not away. Or maybe instead of the new `bool` argument we could add a new option to `StatsDConfiguration` that defaults to preferring buffering? Then we don't need to change the existing signatures at all.
  1. Bump the version to `3.2.0`.

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -1,146 +1,156 @@
...
+                throw new ArgumentNullException(nameof(configuration));
             }
 
-            if (string.IsNullOrWhiteSpace(configuration.Host))
+            var endpointSource = EndpointParser.MakeEndPointSource(
+                configuration.Host, configuration.Port, configuration.DnsLookupInterval);
+
+            var transport = new PooledUdpTransport(endpointSource);
+
+            if (preferBufferedTransport)
```
martincostello: As alluded to in my comment, maybe we could have `if (configuration.PreferBufferedTransport)` instead?

dv00d00: Now it seems so obvious, thx @martincostello

dv00d00:
pushed updates as requested

version.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
-    <VersionPrefix>3.1.1</VersionPrefix>
+    <VersionPrefix>3.2.0</VersionPrefix>
```
martincostello: Can you update `appveyor.yml` as well please?

https://github.com/justeat/JustEat.StatsD/blob/a6a1319586a6545e21e14ffbed466a22ac350620/appveyor.yml#L2

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -1,146 +1,146 @@
...
+            if (transport == null)
+            {
+                throw new ArgumentNullException(nameof(transport));
+            }
 
-            _formatter = new StatsDMessageFormatter(configuration.Prefix);
-            _onError = configuration.OnError;
+            switch (transport)
+            {
+                case IStatsDBufferedTransport transportV2 when configuration.PreferBufferedTransport:
```
martincostello: Now we've dropped V2, maybe this should be called `bufferedTransport`?

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -1,146 +1,146 @@
...
+
     /// <summary>
     ///     Will synchronously publish stats at statsd as you make calls; will not batch sends.
     /// </summary>
     public class StatsDPublisher : IStatsDPublisher
     {
-        private readonly StatsDMessageFormatter _formatter;
-        private readonly IStatsDTransport _transport;
-        private readonly Func<Exception, bool> _onError;
+        private readonly IStatsDPublisher _publisher;
```
martincostello: As this class is now more of a wrapper, maybe call this `_inner`?

martincostello:
I think I'm basically happy with this now as a `3.2.0` release - great job! I'd like @AnthonySteele to review it as well though before we merge it.

Can we get one last set of performance test results from the tip of this branch?

dv00d00:
This run of benchmarks is taken from the different laptop. So could not be compared directly.

`Utf8FormatterBenchmark`

|      Method |     Mean |     Error |   StdDev |   Median | Scaled | ScaledSD |  Gen 0 | Allocated |
|------------ |---------:|----------:|---------:|---------:|-------:|---------:|-------:|----------:|
| StringBased | 876.9 ns | 17.473 ns | 35.69 ns | 862.9 ns |   1.00 |     0.00 | 0.4416 |     696 B |
| BufferBased | 392.1 ns |  7.869 ns | 18.55 ns | 389.7 ns |   0.45 |     0.03 |      - |       0 B |

`UdpStatSendingBenchmark`

|Method |        Mean |      Error |     StdDev | Scaled |  Gen 0 | Allocated |
|---------------------------------- |------------:|-----------:|-----------:|-------:|-------:|----------:|
|                       SendStatUdp | 1,131.69 us | 22.6319 us | 58.0142 us |   1.00 |      - |    1184 B |
|                 SendStatPooledUdp |    31.63 us |  0.6324 us |  1.1240 us |   0.03 | 0.1831 |     384 B |
|         SendStatPooledUdpBuffered |    30.85 us |  0.5901 us |  0.6796 us |   0.03 |      - |       0 B |
| SendStatPooledUdpCoveredByAdapter |    31.96 us |  0.6388 us |  1.6374 us |   0.03 |      - |       0 B |

`StatSendingBenchmark`

|Method |         Mean |      Error |     StdDev |       Median |  Gen 0 | Allocated |
|------------------------ |-------------:|-----------:|-----------:|-------------:|-------:|----------:|
|                  RunUdp | 2,503.119 us | 49.9168 us | 77.7144 us | 2,485.259 us |      - |    2944 B |
|      RunUdpWithSampling |   199.564 us |  3.8886 us |  5.8202 us |   199.324 us | 0.2441 |     597 B |
|                   RunIp | 2,432.112 us | 47.6688 us | 71.3484 us | 2,416.822 us |      - |    2944 B |
|       RunIpWithSampling |   222.738 us |  6.0002 us | 17.1189 us |   223.859 us | 0.2441 |     601 B |
|             RunBuffered |    83.943 us |  2.1686 us |  6.3943 us |    85.700 us |      - |       0 B |
| RunBufferedWithSampling |     6.422 us |  0.1279 us |  0.1617 us |     6.432 us |      - |       0 B |

src/JustEat.StatsD/Buffered/Buffer.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+
+namespace JustEat.StatsD.Buffered
+{
+    internal ref struct Buffer
+    {
+        public Buffer(Span<byte> source) : this()
```
slang25: Last minute thought, what's with `: this()`, isn't that redundant?

dv00d00: thx, good catch

AnthonySteele:
As discussed with marticC

Look good, might need some trials under load.

Not sure if this will be V3.2 with no / minimal breaking changes or 4.0 with other things tidied up as well.

When we can have breaking changes, we don't need both code paths: if we have two classes that work the same, and we know that they both work well in production loads, then we only need one of them (the faster / lower allocation one), there is no real user choice here. 

There still in a choice of Udp vs. Ip transports, we need both for different cases (e.g. AWS lambda). The idea of "a new factory type that could simplify the transport creation and hide the details" sounds worthwhile. So make the built-in cases trivial, and keep the existing more complex ways for users to plug in their own code overrides. I prefer enum options to bool in this case.

dv00d00:
Any updates on this one @martincostello @AnthonySteele ?

martincostello:
I think we just need to decide whether this is a `3.2.0` or a `4.0.0` (as this PR changed the number, if it's going to be a 4.0.0 we should update it here before merging), then this is good to merge.

AnthonySteele:
If we go for a `4.0.0` then merge this and allow other changes before finalising it.

martincostello:
Myself and @AnthonySteele discussed this offline, and we're going to release this as a `3.2.0` later today as something users can opt-in to if they want to use it. Then we'll make this and the changes from #81 the defaults and no longer an opt-in as part of a future `4.0.0` release.

PR #107, opened 2018-10-02
Delete unused PoolAwareSocketAsyncEventArgs

AnthonySteele:

Delete the class `PoolAwareSocketAsyncEventArgs`. It is no longer used in this library. It is only still here due to circumstances. 

This should go into the next major release.: Technically it is a breaking change (that's why it has been held off until now).  But it is unlikely to actually affect most users. If you somehow are using  `PoolAwareSocketAsyncEventArgs`, you should take a copy of the source into your code.


martincostello:
Suggest that #108 goes in before this one so that this change has the right version when it builds in master.

PR #109, opened 2018-10-02
Simpler func for SimpleObjectPool

AnthonySteele:

Simpler func for SimpleObjectPool

For the SimpleObjectPool constructor:
- The constructor func can be `Func<T>`, we don't use the other param at all
- the constructor func is IMHO a more important param than an initial size, so should come first
- initial size has a default of zero


AnthonySteele:
todo in a later PR:  `SimpleObjectPool<T>` is only ever used as a `SimpleObjectPool<Socket>` - is the generic necessary?

`ConnectedSocketPool` is a wrapper of `SimpleObjectPool<Socket>`, which means that each operation is forwarded, incurring a tiny extra delay. Fold these classes together?

src/JustEat.StatsD/SimpleObjectPool.cs
```diff
@@ -8,16 +8,16 @@ namespace JustEat.StatsD
     /// </summary>
     public sealed class SimpleObjectPool<T> where T : class
```
martincostello: We can make this `internal` now too.

martincostello:
And yes you're right, the generic is also redundant now too.

src/JustEat.StatsD/SimpleObjectPool.cs
AnthonySteele: Good point, done

AnthonySteele:
I plan to make a later PR to fold `SimpleObjectPool` into `ConnectedSocketPool` as this is the only place where it is used. It will be simpler, and _might_ remove some small run-time overhead - the indirection might or might not be optimised out.

As far as I know, there is no suitable generic pool class in the framework, as we are already doing [something very close to this example](https://docs.microsoft.com/en-us/dotnet/standard/collections/thread-safe/how-to-create-an-object-pool)

PR #112, opened 2018-10-03
when formatting timing use miliseconds instead of ticks

### LINKED ISSUE #110: Incorrect timing stats published in 3.2.0
### Expected behaviour

Timings are consistent with previous version (`3.1.1`) in an otherwise like-for-like environment.

### Actual behaviour

Stats are incorrect by an order of magnitude, giving timings in hours which are impossible.

image

### Steps to reproduce

Generate timing metrics using `3.2.0`.

/cc @dv00d00

dv00d00:
Timing was using ticks instead of milliseconds :sigh:

Fixes #110


martincostello:
Looks like I'll need to cherry-pick this into a branch from master once this is merged so we can publish a fix.

PR #113, opened 2018-10-03
Fold the SimpleObjectPool into ConnectedSocketPool

AnthonySteele:

Fold the `SimpleObjectPool` into `ConnectedSocketPool`, as that is the only place that it used.

Follow on from #109

Todo in later PRs:
`ConnectedSocketPool` is used in `PooledUdpTransport`.
We also have:
*  `UdpTransport`  - the unpooled version, suggest that that one is deleted in `4.0.0` 
* `IpTransport` which is useful for AWS Lambdas, but could also benefit from pooling? `ConnectedSocketPool` could also be extended to have option to store the IP sockets if need be?


src/JustEat.StatsD/ConnectedSocketPool.cs
```diff
@@ -7,22 +8,35 @@ namespace JustEat.StatsD
     internal sealed class ConnectedSocketPool : IDisposable
     {
         private bool _disposed;
-        private readonly SimpleObjectPool<Socket> _pool;
+        private readonly ConcurrentBag<Socket> _pool = new ConcurrentBag<Socket>();
+
+        public IPEndPoint IpEndPoint { get; }
 
         public ConnectedSocketPool(IPEndPoint ipEndPoint)
```
martincostello: Maybe we should make `count` a parameter and have the caller specify the value? (which would happen to be `Environment.ProcessorCount`)

src/JustEat.StatsD/ConnectedSocketPool.cs
```diff
@@ -7,22 +8,35 @@ namespace JustEat.StatsD
...
+            {
+                _pool.Add(CreateSocket());
+            }
         }
 
-        private static Socket CreateSocket(IPEndPoint ipEndPoint)
+        private Socket CreateSocket()
         {
             var socket = UdpTransport.CreateSocket();
+            if (socket == null)
```
martincostello: [`UdpTransport.CreateSocket()`](https://github.com/justeat/JustEat.StatsD/blob/b1edcccf63797a73c777d606ab406a989a2f6fdb/src/JustEat.StatsD/UdpTransport.cs#L62-L77) never returns `null`.

PR #114, opened 2018-10-04
remove the unpooled UdpTransport

AnthonySteele:

remove the unpooled `UdpTransport` and keep the `PooledUdpTransport`
as pooled is preferred and we don't need 2 ways to do the same thing.

Follow on from #113


src/JustEat.StatsD/Transport.cs
```diff
@@ -0,0 +1,30 @@
+using System.Net.Sockets;
+using System.Runtime.InteropServices;
+
+namespace JustEat.StatsD
+{
+    internal static class Transport
```
martincostello: `SocketFactory`?

src/JustEat.StatsD/Transport.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        internal static Socket UdpSocket()
```
martincostello: `ForUdp()`?

src/JustEat.StatsD/Transport.cs
```diff
@@ -0,0 +1,30 @@
...
+                socket.SendBufferSize = 0;
+            }
+#else
+            socket.SendBufferSize = 0;
+#endif
+
+            return socket;
+        }
+
+        internal static Socket IpSocket()
```
martincostello: `ForIP()`?

martincostello:
Suggest in a PR after this one to rename `PooledUdpTransport` to `UdpTransport` - the pooling is now just an implementation detail, so I don't think it needs to be in the type name.

src/JustEat.StatsD/Transport.cs
AnthonySteele: I have changed all 3 as per comments

PR #115, opened 2018-10-04
Rename pooled

AnthonySteele:

Rename `PooledUdpTransport` to `UdpTransport` as it's the only Udp transport.
Rename vars, tests, benchmarks to match.

Follow-on from #114

### DIFF
--- a/src/Benchmark/UdpStatSendingBenchmark.cs
+++ b/src/Benchmark/UdpStatSendingBenchmark.cs
@@ -11,7 +11,7 @@ namespace Benchmark
     {
         private static readonly TimeSpan Timed = TimeSpan.FromMinutes(1);
 
-        private PooledUdpTransport _pooledTransport;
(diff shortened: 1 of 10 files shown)


src/Benchmark/UdpStatSendingBenchmark.cs
```diff
@@ -21,7 +21,7 @@ public void Setup()
         {
             var config = new StatsDConfiguration
             {
-                // if you want verify that stats are recevied,
+                // if you want verify that stats are received,
```
martincostello: Maybe do a find-and-replace and see if there's any more of these as it's the second one today? 😃

AnthonySteele: I've checked, there are no more.  But I changed the wording anyway; "you want verify" was not quite right.

PR #116, opened 2018-10-05
Easy set transport

AnthonySteele:

The `UDPTransport` is far ahead of the `IPTransport` and I think that this is not an essential thing, simply due to all the development effort being focussed on the more common UDP scenario. It seems that the only _essential_ way that these differ is in the variety of socket created.

So  `UDPTransport` becomes `SocketTransport` that can create either kind of socket, controlled by a 2-state enum `SocketProtocol`.

`IPTransport` or something to replace it is necessary: We still need to make IP sockets for AWS Lambda cases, but there should be an easier way to do that common case.  A `SocketProtocol` flag on the configuration solves that.

The more complex custom setup should still be just as possible as before.

We will need to test this with both kinds of socket, localy and then in Lambdas with IP sockets, and in regular features with UDP.


src/Benchmark/StatSendingBenchmark.cs
```diff
@@ -28,11 +28,11 @@ public void Setup()
             var endpointSource = EndpointParser.MakeEndPointSource(
                 config.Host, config.Port, config.DnsLookupInterval);
 
-            var ipTransport = new IpTransport(endpointSource);
+            var ipTransport = new SocketTransport(endpointSource, SocketProtocol.Ip);
```
martincostello: I _think_ that naming guidelines is that for acronyms of two letters you capitalise, so `IP` rather than `Ip`.

src/Benchmark/UdpStatSendingBenchmark.cs
```diff
@@ -32,7 +32,7 @@ public void Setup()
                 config.Port,
                 config.DnsLookupInterval);
 
-            _transport = new UdpTransport(endpointSource);
+            _transport = new SocketTransport(endpointSource, SocketProtocol.Udp);
```
martincostello: Is there any utility in making an overload that just takes the IP endpoint and defaults to `Udp` for the common case?

src/JustEat.StatsD/SocketProtocol.cs
```diff
@@ -0,0 +1,8 @@
+namespace JustEat.StatsD
+{
+    public enum SocketProtocol
```
martincostello: `///` docs.

src/JustEat.StatsD/SocketTransport.cs
```diff
@@ -9,23 +9,27 @@ namespace JustEat.StatsD
...
         /// <param name="endPointSource">The <see cref="IPEndPointSource"/> to use.</param>
+        /// <param name="socketProtocol">Udp or Ip sockets</param>
         /// <exception cref="ArgumentNullException">
         /// <paramref name="endPointSource"/> is <see langword="null"/>.
         /// </exception>
-        public UdpTransport(IPEndPointSource endPointSource)
+        public SocketTransport(IPEndPointSource endPointSource, SocketProtocol socketProtocol)
         {
             _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
+            _socketProtocol = socketProtocol;
```
martincostello: Should we validate the value is valid here, rather than waiting for it to be thrown later when used to create a socket?

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -44,10 +43,15 @@ public class StatsDConfiguration
 
         /// <summary>
         /// If true will prefer to use <see cref="IStatsDBufferedTransport"/> interface implementation
-        /// if transport implementation provides both 
+        /// if transport  implementation provides both
```
martincostello: Accidentally extra space?

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -44,10 +43,15 @@ public class StatsDConfiguration
...
         /// If true will prefer to use <see cref="IStatsDBufferedTransport"/> interface implementation
-        /// if transport implementation provides both 
+        /// if transport  implementation provides both 
         /// <see cref="IStatsDBufferedTransport"/> and <see cref="IStatsDTransport"/> interfaces.
         /// Default is true.
         /// </summary>
         public bool PreferBufferedTransport { get; set; } = true;
+
+        /// <summary>
+        /// Configure to use either UDP or IP sockets to transport stats. Defaults to UDP
```
martincostello: Full stop at end.

src/Benchmark/StatSendingBenchmark.cs
AnthonySteele: correct, "capitalizes the first character of each word (including acronyms over two letters in length),"

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

src/Benchmark/UdpStatSendingBenchmark.cs
AnthonySteele: I don't see a need for it, I would not recommend using that internally

src/JustEat.StatsD/SocketProtocol.cs
martincostello: Need docs for the values too.

src/JustEat.StatsD/SocketProtocol.cs
```diff
@@ -1,8 +1,12 @@
 namespace JustEat.StatsD
 {
+    /// <summary>
+    /// The subset of ProtocolType that are supported by SocketTransport
+    /// UDP or IP
```
martincostello: Full stop.

src/JustEat.StatsD/SocketProtocol.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// The subset of ProtocolType that are supported by SocketTransport
+    /// UDP or IP.
+    /// UDP is the default, but IP transport is required for AWS Lambdas.
+    /// </summary>
+    public enum SocketProtocol
+    {
+        /// <summary>
+        /// transport stats over UDP sockets
```
martincostello: _"Transport stats over UDP sockets."_

src/JustEat.StatsD/SocketProtocol.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// </summary>
+        Udp,
+
+        /// <summary>
+        /// transport stats over IP sockets
```
martincostello: _"Transport stats over IP sockets."_

AnthonySteele:
I found that `SocketProtocol` enum hard to name - It can't be `SocketTransport`. I am open to other name suggestions.

martincostello:
`SocketMode`?

src/JustEat.StatsD/SocketTransport.cs
AnthonySteele: IMHO not. It's customary to always have a `default` case on a switch on an enum so that it's exhaustive, and never returns null in a quiet failure. You would have to try hard to get an enum value that's not one of the two valid ones.

martincostello: We're validating it in `SocketFactory` though - it's better to throw an error upfront on construction by validating the parameters, than leaving it until later.

src/JustEat.StatsD/SocketTransport.cs
```diff
@@ -29,6 +29,11 @@ public sealed class SocketTransport : IStatsDTransport, IStatsDBufferedTransport
         public SocketTransport(IPEndPointSource endPointSource, SocketProtocol socketProtocol)
         {
             _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
+
+            if (socketProtocol != SocketProtocol.IP && socketProtocol != SocketProtocol.Udp)
+            {
+                throw new ArgumentOutOfRangeException(nameof(socketProtocol));
```
martincostello: If you use [this constructor](https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception.-ctor?view=netframework-4.7.2#System_ArgumentOutOfRangeException__ctor_System_String_System_Object_System_String_) the offending value can be included to improve the debugging experience.

src/JustEat.StatsD/SocketTransport.cs
```diff
@@ -29,6 +29,11 @@ public sealed class SocketTransport : IStatsDTransport, IStatsDBufferedTransport
         public SocketTransport(IPEndPointSource endPointSource, SocketProtocol socketProtocol)
         {
             _endpointSource = endPointSource ?? throw new ArgumentNullException(nameof(endPointSource));
+
+            if (socketProtocol != SocketProtocol.IP && socketProtocol != SocketProtocol.Udp)
```
martincostello: If you use [`Enum.IsDefined(Type, object)`](https://docs.microsoft.com/en-gb/dotnet/api/system.enum.isdefined?view=netframework-4.7.2#System_Enum_IsDefined_System_Type_System_Object_) then we don't need to change this code if we add a new value to the enum.

src/JustEat.StatsD/SocketTransport.cs
slang25: Amen to that @AnthonySteele. One downside is that `Enum.IsDefined` allocates, although if this class is pooled then it doesn't matter. They aren't going to make runtime changes to remove the allocation (Like they did with `Enum.GetHastCode` in .NET Core 2.0), so we'd need to wait on the enum generic type constraint to land in C# and for a generic version of `Enum.IsDefined` to land.

martincostello: I think based on the use case, the allocation isn't really a concern as you wouldn't expect lots of transports to be being created.

PR #117, opened 2018-10-06
Remove the "PreferBufferedTransport" flag

AnthonySteele:

The "PreferBufferedTransport" flag should always be true. In Version 4.0 we are are removing redundant code paths because 
- we can make breaking changes (although the majority of the consumers should be able to just recompile and carry on) 
- and these "experiments" can be productionised now.  We keep the best/fastest path and drop the rest. The new code is ether ready for mainstream use, or should be made so.

So should the `StatsDPublisher` with an `IStatsDPublisher _inner` also go?
And the distinction between `IStatsDTransport` and `IStatsDBufferedTransport` ?

### DIFF
--- a/src/JustEat.StatsD.Tests/WhenCreatingStatsDPublisher.cs
+++ b/src/JustEat.StatsD.Tests/WhenCreatingStatsDPublisher.cs
@@ -77,8 +77,7 @@ namespace JustEat.StatsD
             var config = new StatsDConfiguration
             {
                 Prefix = "test",
-                Host = "127.0.0.1",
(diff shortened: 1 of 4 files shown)


martincostello:
Have added some musing on the above in #118.

martincostello:
So do you reckon this is ready to merge as-is, then I'll rebase #118 and merge that?

We can then do any allocation-savings once we've consolidated things a bit.

AnthonySteele:
If you want, we can merge this as is, and then consolidate after.
That bool `PreferBufferedTransport` should go, one way or the other.

PR #118, opened 2018-10-06
Merge IStatsDTransport with IStatsDBufferedTransport

martincostello:
Merge `IStatsDBufferedTransport` with `IStatsDTransport` by renaming the former to the latter and then removing the latter so there's only one transport interface.

An implicit conversion from one to the other is created by adding an extension method to converts `string` to `ArraySegment<byte>` which makes the change source compatible.

Relates to #117 and would follow that after a rebase if we chose to go down this route.

This isn't as performant in the case that someone uses the transport themselves directly using strings as the optimisations that `SocketTransport` had for the `Send(string)` to skip the conversion to `ArraySegment<T>` and send the `byte[]` directly to the socket aren't available (see below), but I'm not sure whether that's a problem as it's source compatible, and if you _really_ wanted to do some low-level stuff not through the publisher, I guess you'd probably write your code to not use strings anyway.

https://github.com/justeat/JustEat.StatsD/blob/3ca60f3cdac2d9cfde4f66c73c19de35dd1c8c15/src/JustEat.StatsD/SocketTransport.cs#L41-L63


AnthonySteele:
`ArrraySegment<T>` seems to be [available in .Net framework 2.0 onwards according to the docs  ](https://docs.microsoft.com/en-us/dotnet/api/system.arraysegment-1)

As long as that code path via `BufferBasedStatsDPublisher.SendMessage` is fast and low allocation in e.g. `NET4.6.1`  and similar, then we should do this.

martincostello:
I've rebased this and also now removed `StringBasedStatsDPublisher` too. There's probably a few more bits of cruft lying around in the tests/benchmarks, but we can tidy them up later as we find them.

martincostello:
@AnthonySteele If you could re-review this since my rebase, I'll merge this in if you're happy.

src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -73,6 +65,11 @@ public StatsDPublisher(StatsDConfiguration configuration)
                 throw new ArgumentNullException(nameof(configuration));
             }
 
+            if (string.IsNullOrWhiteSpace(configuration.Host))
+            {
+                throw new ArgumentException("No hostname is set.", nameof(configuration));
```
AnthonySteele: technically this a "host name or IP", and usually an IP not a name.

PR #122, opened 2018-10-06
Add integration test

martincostello:
Add an integration test that sends metrics to a real statsd instance running locally ([docs](https://github.com/etsy/statsd#installation-and-configuration)) in AppVeyor or Travis, and verifies that the values sent to the server are correct by using the admin endpoint on port `8126` ([docs](https://github.com/etsy/statsd/blob/master/docs/admin_interface.md#statsd-specific-commands)).

A unique random prefix is used so that if you run these tests locally the values don't clash.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,9 @@ addons:
     - libssl-dev
     - libunwind8
 
+install:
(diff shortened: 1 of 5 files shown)


AnthonySteele:
Nice!

PR #125, opened 2018-10-07
Implement IDisposable on StatsDPublisher

### LINKED ISSUE #119: StatsDPublisher(StatDConfiguration) creates a resource leak
The `StatsDPublisher(StatsDConfiguration)` constructor creates a resource leak.

https://github.com/justeat/JustEat.StatsD/blob/3ca60f3cdac2d9cfde4f66c73c19de35dd1c8c15/src/JustEat.StatsD/StatsDPublisher.cs#L39-L59

Given that `SocketTransport` implements `IDisposable`, using this constructor will leak sockets beyond the lifetime of the publisher as there is no way to deterministically dispose of the sockets it pools.

This is only really a problem where the publisher isn't a singleton and/or isn't used for the lifetime of the application.

martincostello:
* Make the `StatsDPublisher` class `IDisposable` to fix #119.
  * Change `StatsDPublisher` to be `sealed` as there are no useful extension points and would otherwise require a finalizer and a `protected virtual void Dispose(bool disposing)` method for `IDisposable` implementation for any derived classes.
  * Remove redundant ReSharper warning comment from #104.
  * Remove some redundant null checks on some internal types.
  * Add XML `///` documentation to `StatsDPublisher`.

### DIFF
--- a/src/Benchmark/UdpStatSendingBenchmark.cs
+++ b/src/Benchmark/UdpStatSendingBenchmark.cs
@@ -14,7 +14,7 @@ namespace Benchmark
         private SocketTransport _transport;
         private StringBasedStatsDPublisher _udpSender;
         private BufferBasedStatsDPublisher _bufferBasedStatsDPublisher;
-        private IStatsDPublisher _adaptedStatsDPublisher;
+        private StatsDPublisher _adaptedStatsDPublisher;
 
         [GlobalSetup]
         public void Setup()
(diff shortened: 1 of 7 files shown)


martincostello:
I think that’s fair. Also once the churn settles down, I want to go through all the public members and add proper `///` docs and ensure we ship them with the NuGet package so we light up Intellisense for IDEs that support it.

martincostello:
Have added #130 to track any docs updates.

PR #126, opened 2018-10-07
Add static analyzers

martincostello:
This PR adds Roslyn code analyzers (a bit like FxCop of old) that run at build time. It also fixes various issues identified by them including:
  1. Redundant methods caused by changes to `IStatsDPublisher` over time.
  1. Changes methods that don't use instance data to be static.
  1. Makes the internal `DisposableTimer` class sealed.
  1. Fixes an incorrectly setup exception for parameter validation.
  1. Uses `Array.Empty<T>()` to save an allocation in target frameworks that support it.
  1. Specifies `ConfigureAwait(false)` when awaiting tasks.

### DIFF
--- a/src/JustEat.StatsD.Tests/SocketTransportTests.cs
+++ b/src/JustEat.StatsD.Tests/SocketTransportTests.cs
@@ -6,7 +6,7 @@ using Xunit;
 
 namespace JustEat.StatsD
 {
-    public class SocketTransportTests
+    public static class SocketTransportTests
     {
         [Fact]
         public static void ValidSocketTransportCanBeConstructed()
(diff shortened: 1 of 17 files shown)


martincostello:
This PR will probably need rebasing after #117, #118 and #125.

src/JustEat.StatsD.Tests/SocketTransportTests.cs
```diff
@@ -35,16 +35,21 @@ public static void SocketTransportCanSendOverIPWithoutError()
...
                 () => new SocketTransport(null, SocketProtocol.IP));
-
         }
 
         [Fact]
         public static void InvalidSocketProtocolThrowsInConstructor()
         {
-            Should.Throw<ArgumentOutOfRangeException>(
-                () => new SocketTransport(LocalStatsEndpoint(), (SocketProtocol)42));
+            SocketProtocol socketProtocol = (SocketProtocol)42;
```
AnthonySteele: Can you use `var` here? and name it `invalidSocketProtocol` ?

martincostello: `var` is fine, but I prefer to have the variable and argument names match for simple tests scenarios like this. Happy to change it if we really want the name like that though.

PR #127, opened 2018-10-07
More integration tests

martincostello:
1. Add integration tests for more scenarios, such as hostname vs. IP address, buffered vs. not and IP vs. UDP.
  1. Put the returned JSON into the assert messages so they can be viewed if the test fails.
  1. If a real statsd service is running on port 8125 on localhost, it logged errors complaining about "testStat".
  1. Fix the asserts sometimes querying the management interface before stats would be visible.
  1. Increase the statsd flush interval so that metrics don't get removed before we can assert on them as the default was 10 seconds.

~~This PR has highlighted that the tests from #122 appear to be slightly flaky, with statsd not reporting the expected metrics as being received. Needs further investigation as to _why_.~~

~~This will also need rebasing if merged after #117 and/or #118 or #125 as non-buffered won't be a thing anymore and/or will need to account for `IDisposable`~~.

### DIFF
--- a/src/JustEat.StatsD.Tests/IntegrationTests.cs
+++ b/src/JustEat.StatsD.Tests/IntegrationTests.cs
@@ -10,71 +10,89 @@ namespace JustEat.StatsD
 {
     public static class IntegrationTests
     {
-        [SkippableFact]
-        public static async Task Can_Send_Metrics_To_StatsD()
+        [SkippableTheory]
+        [InlineData("localhost", SocketProtocol.IP, false)]
+        [InlineData("localhost", SocketProtocol.Udp, false)]
+        [InlineData("127.0.0.1", SocketProtocol.IP, false)]
+        [InlineData("127.0.0.1", SocketProtocol.Udp, false)]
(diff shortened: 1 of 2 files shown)


martincostello:
Depends on #128.

martincostello:
[This happens](https://travis-ci.org/justeat/JustEat.StatsD/jobs/438321913#L360) occasionally, I wonder if there's a bug in the `TcpClient` usage that makes the node server hosting statsd log a warning.

```sh
7 Oct 17:36:48 - WARNING: MGMT: Caught Error: read ECONNRESET, Moving on
```

martincostello:
I think this is ready to go as soon as we merge #118.

martincostello:
The following commits are also candidates to go to `release/3xx` to fix test flakiness:
  1. b98ef71f97162c65d67e47bbc9ace672744fbdf7
  1. 725b9db411bd7a23a767abcb6bba7e859de42c50
  1. a140c7cd5899341dafa7d642ca83b5dc9ae64520

martincostello:
This should be good to go now, then I think the four combinations are enough to validate things at an integration level going forward.

PR #124, opened 2018-10-07
Sync fixes to release/3xx

martincostello:
Merge the fix from #121 and the tests from #122 into the `release/3xx` branch via cherry-picking so that if we need to do another `3.x.x` release, we've got some safe changes from master in that branch.

Also bumps the version in this branch to `3.2.2`.

Replaces #123.

PR #128, opened 2018-10-07
Fix unbuffered batch metrics send

martincostello:
When I added more integration tests in #127, it found a bug where multiple metrics being sent at once using the "old" implementation (the existing one at the time of #104) would be malformed and dropped when sent to statsd.

This PR fixes that by making the following changes and optimisations uncovered while debugging the failing tests:
  1. Fix malformed metrics by sending multiple metrics as different sends, not one send with everything concatenated together. It may be that this is allowed, but I couldn't find what the delimiter would be in that scenario, so I just switched to a `foreach` liked the buffered implementation does.
  1. Add optimisations to not send empty strings to statsd if the sample rate means that none of the metrics need to be sent.
  1. Add null and zero check for buckets array to fix potential for a `NullReferenceException` as well as removing the need to allocate an enumerator.

I guess that we should include this with #124 and do a `3.2.2` release with the fix?

### DIFF
--- a/src/JustEat.StatsD.Tests/StringBasedStatsDPublisherTests.cs
+++ b/src/JustEat.StatsD.Tests/StringBasedStatsDPublisherTests.cs
@@ -1,0 +1,89 @@
+using System;
+using Moq;
+using Xunit;
+
+namespace JustEat.StatsD
+{
+    public static class StringBasedStatsDPublisherTests
+    {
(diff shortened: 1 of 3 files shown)


martincostello:
We should merge this before #115 or #118 so it's easy to cherry-pick into the `release/3xx` branch.

PR #132, opened 2018-10-08
Quick review

AnthonySteele:

Two things from review that are not vital, but are nice to have
-  don't keep a reference to a disposed `_pool`
- `_disposeTransport` is only set in the constructor, can be `readonly`.

### DIFF
--- a/src/JustEat.StatsD/StatsDPublisher.cs
+++ b/src/JustEat.StatsD/StatsDPublisher.cs
@@ -17,7 +17,7 @@ namespace JustEat.StatsD
         private readonly IStatsDTransport _transport;
 
         private bool _disposed;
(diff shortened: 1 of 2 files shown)


src/JustEat.StatsD/StatsDPublisher.cs
```diff
@@ -17,7 +17,7 @@ public sealed class StatsDPublisher : IStatsDPublisher, IDisposable
         private readonly IStatsDTransport _transport;
 
         private bool _disposed;
-        private bool _disposeTransport;
+        private readonly bool _disposeTransport;
```
martincostello: I've covered this one as part of the refactor in #118.

PR #136, opened 2018-10-10
Base Endpoint

AnthonySteele:

We can use the base class `EndPoint` not `IpEndpoint` in quite a few places, since that's what we connect the socket to.

When we actually make an endpoint, we make the subclass, `IPEndpoint`.

### DIFF
--- a/src/Benchmark/UdpTransportBenchmark.cs
+++ b/src/Benchmark/UdpTransportBenchmark.cs
@@ -14,12 +14,12 @@ namespace Benchmark
         private SocketTransport _transport;
         private SocketTransport _transportSwitched;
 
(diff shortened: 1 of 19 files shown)


martincostello:
J'approve.

src/Benchmark/UdpTransportBenchmark.cs
```diff
@@ -14,18 +14,18 @@ public class UdpTransportBenchmark
         private SocketTransport _transport;
         private SocketTransport _transportSwitched;
 
-        private class MilisecondSwitcher : IPEndPointSource
+        private class MilisecondSwitcher : IEndPointSource
```
shaynevanasperen: This class name is spelled wrong. It should have two L's.

PR #137, opened 2018-10-10
Simplify IStatsDPublisher

martincostello:
* Simplify the `IStatsDPublisher` interface by removing all the overloads and moving them to convenience extension methods.
  * Remove the `MarkEvent()` method as it's just an alias to `Increment()`.
  * Add XML documentation to `IStatsDPublisher`.

### DIFF
--- a/src/Benchmark/StatSendingBenchmark.cs
+++ b/src/Benchmark/StatSendingBenchmark.cs
@@ -41,7 +41,7 @@ namespace Benchmark
         [Benchmark]
         public void RunIp()
         {
-            _ipSender.MarkEvent("hello.i");
+            _ipSender.Increment("hello.i");
             _ipSender.Increment(20, "increment.i");
             _ipSender.Timing(Timed, "timer.i");
             _ipSender.Gauge(354654, "gauge.i");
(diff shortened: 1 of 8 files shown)


AnthonySteele:
`MarkEvent` is used in our code. The intent is to "increment" something that happens quite rarely, e.g. new version of the app starts up. A vertical line that that point helps correlate this "event" with changes to other stats. 

Removing it it seems a slightly different concern to the rest of this PR. but it can be yet another extension method, as it's really just an increment.

martincostello:
`MarkEvent(string)` had _exactly_ the same implementation as `Increment(string)` in terms of how it worked, which I found when I started to create the extension method to implement it. Given that it was just copy-paste duplication for the sake of it to create an alias, it went on the cutting room floor.

It adds no value to this library, so I'd argue that if it's something you find semantically valuable yourself, you add an extension method to create the alias to your own application(s).

To that end, it comes under the purview of _"simplification"_, which is why it's in this PR.

AnthonySteele:
We would have to document this in the release, that `MarkEvent` can be replaced with `Increment` throughout.

martincostello:
That's fine, it can go in the release notes with all the other breaking changes we've made.

Issue #130, opened 2018-10-07
Documentation review


martincostello:
As discussed in https://github.com/justeat/JustEat.StatsD/pull/125#issuecomment-427665559, we should review the documentation, both in code and elsewhere (such as the `README`) for the `4.0.0` release.

martincostello:
We should also standardise our naming so we always call it _"StatsD"_ (not _"statsd"_ or _"statsD"_).

PR #139, opened 2018-10-10
XML documentation updates

martincostello:
* Add missing XML documentation.
  * Add some missing parameter validation.
  * Updates to grammar in some error messages.
  * Fix some incorrect XML documentation.
  * Make `StatsDMessageFormatter` internal (so we don't have to document it).
  * Add some extra overloads to `TimerExtensions` to remove the need to have a parameter for `IDisposableTimer` if you aren't going to use it, which can help save on allocations for delegates if you don't otherwise need a lambda.

Relates to #130.

### DIFF
--- a/src/JustEat.StatsD/IDisposableTimer.cs
+++ b/src/JustEat.StatsD/IDisposableTimer.cs
@@ -1,9 +1,15 @@
-﻿using System;
+using System;
 
 namespace JustEat.StatsD
 {
+    /// <summary>
+    /// Defines a timer which is disposable.
+    /// </summary>
     public interface IDisposableTimer : IDisposable
     {
+        /// <summary>
+        /// Gets the name of the statsd bucket associated with the timer.
+        /// </summary>
         string StatName { get; set; }
     }
 }
--- a/src/JustEat.StatsD/IDisposableTimer.cs
+++ b/src/JustEat.StatsD/IDisposableTimer.cs
@@ -8,7 +8,7 @@ namespace JustEat.StatsD
     public interface IDisposableTimer : IDisposable
     {
         /// <summary>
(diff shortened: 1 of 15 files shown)


martincostello:
The idea behind the new timer overloads is so where before you did this:

```csharp
public void MyMethod()
{
    // Whatever
]

public void MyOtherMethod()
{
    publisher.Time("mymethod", _ () => MyMethod());
}
```

Now you can just do this

```csharp
public void MyMethod()
{
    // Whatever
]

public void MyOtherMethod()
{
    publisher.Time("mymethod", MyMethod);
}
```

This removes the need for the compiler to allocate extra delegates to deal with the anonymous lambda required to deal with the `IDisposableTimer` parameter.

src/JustEat.StatsD/IDisposableTimer.cs
```diff
@@ -1,9 +1,15 @@
...
 
 namespace JustEat.StatsD
 {
+    /// <summary>
+    /// Defines a timer which is disposable.
+    /// </summary>
     public interface IDisposableTimer : IDisposable
     {
+        /// <summary>
+        /// Gets the name of the statsd bucket associated with the timer.
```
AnthonySteele: "Gets or sets the name" ? It's writable so that you can alter the stat name based on eg. http response code inside the block being timed.

AnthonySteele:
The new overloads seem fine.  I left one comment. The rest is good too!

src/JustEat.StatsD/IDisposableTimer.cs
martincostello: Good spot.

martincostello: One day I'll convince people StyleCop's a good idea, and that would've found this for me.

PR #144, opened 2018-10-13
Add NuGet config

martincostello:
I've been doing some benchmarks for all the releases of JustEat.StatsD since we added .NET Core support to get some comparisons, and that required using a custom NuGet.config to restore packages from the AppVeyor CI feed.

This PR adds the vanilla NuGet.config file to the repo so that's easier to do in future without having to create one from scratch.

I've also extended the BenchmarkDotNet artifact wildcard in `.gitignore` to ignore the slightly different naming convention I'm using locally to generate a folder per StatsD-version/TFM combination.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,4 @@ project.lock.json
 TestResult.xml
 !3rdparty/*/bin
 lint.db
-BenchmarkDotNet.Artifacts
+BenchmarkDotNet.Artifacts*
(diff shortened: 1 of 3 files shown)


slang25:
Another option is to use `RestoreAdditionalProjectSources` in your csproj or Directory.Build.props, then you could put it behind a condition and hit it by passing a command line argument.

```xml
<RestoreAdditionalProjectSources>
    https://yourown.myget.org/...
</RestoreAdditionalProjectSources>
```

AnthonySteele:
It's a bit odd to add a `NuGet.config` that just restates the default value, but if it's going to be used in future,  then we can do that.

PR #145, opened 2018-10-14
Fix NullReferenceException

martincostello:
Whether it's worth publishing a `3.2.3` with _just_ this fix in it or not is debatable, but the `3.x.x` code base has a bug in it where a `NullReferenceException` is thrown on the finalizer thread and terminates the process if an exception occurs while creating a `ConnectedSocketPool` because `_pool` never ends up with a value.

```
The active test run was aborted. Reason: Unhandled Exception: System.NullReferenceException: Object reference not set to
 an instance of an object.
   at JustEat.StatsD.ConnectedSocketPool.Dispose(Boolean disposing) in C:\Coding\JustEat.StatsD\src\JustEat.StatsD\Conne
ctedSocketPool.cs:line 67
   at JustEat.StatsD.ConnectedSocketPool.Finalize() in C:\Coding\JustEat.StatsD\src\JustEat.StatsD\ConnectedSocketPool.c
s:line 51
```

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,5 @@
 os: Visual Studio 2017
-version: 3.2.2.{build}
+version: 3.2.3.{build}
 configuration: Release
 environment:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
--- a/src/JustEat.StatsD.Tests/PooledUdpTransportTests.cs
+++ b/src/JustEat.StatsD.Tests/PooledUdpTransportTests.cs
@@ -1,0 +1,41 @@
+using System;
+using Xunit;
+
+namespace JustEat.StatsD
+{
+    public static class PooledUdpTransportTests
+    {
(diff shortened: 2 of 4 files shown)


martincostello:
This got flushed out by doing some performance testing with version `3.1.1` where the benchmarks didn't work with a value of `localhost` for the host because of the bug fixed by #121.

AnthonySteele:
Hmm, that should not/does not happen commonly in practice. So its definitely worth merging, but I don't feel the need for a `3.x` release _just_ for this change. 

The equivalent should also be on the `master` branch?

martincostello:
This bug can't be replicated in `master` due to the fact the pool is initialized before the constructor runs:

https://github.com/justeat/JustEat.StatsD/blob/7ffb9099abf43a3d43e4df794a5cc71351791642/src/JustEat.StatsD/ConnectedSocketPool.cs#L11

PR #153, opened 2018-12-13
review docs

AnthonySteele:

review of the `Readme.md` document. Needed for a 4.0 release

 related to #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,18 +11,30 @@
 
 ### TL;DR
 
-We use this within our components to publish [StatsD](http://github.com/etsy/statsd) metrics from .NET code. We've been using this in most of our things, since around 2013.
+We use this library within our components to publish [StatsD](http://github.com/etsy/statsd) metrics from .NET code to a server. We've been using this in most of our things, since around 2013.
+
+### Supported platforms
+
+`JustEat.Statsd` version 4 is built for these target frameworks: `net451` , `netstandard2.0`, `netcoreapp2.1`, `netcoreapp2.2`.
 
 ### Features
 
-* StatsD metrics formatter
-* UDP client handling
+* Easy to use
+* Robust and proven
+* Tuned for high performance and low resource usage
+* Works well with modern .NET apps - .NET Core, .NET Standard 2,0 and `async ... await`
+* Supports standard statsd primitives: `Increment`, `Decrement`, `Timing` and `Gauge`. Supports sample rate.
+* Helpers to make it easy to time a `Func`, `Action` or a code block inside a `using` statement.
+* Send stats over UDP or IP
+* Send stats to a server byt name or IP address
 
 #### Publishing statistics
 
-`IStatsDPublisher` is the interface that you will use in most circumstances. With this you can `Increment` or `Decrement` an event, and send values for a `Gauge` or `Timing`.
+`IStatsDPublisher` is the interface that you will use in most circumstances. With this you can `Increment` an event, and send values for a `Gauge` or `Timing`.
 
-The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The constructor takes an instance of `StatsDConfiguration`. For the configuration's values, you will always need the StatsD server host name or IP address. Optionally, you can change the standard port (8125). You can also prepend a prefix to all stats. These values often come from configuration as the host name and/or prefix may vary between test and production environments.
+The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The constructor takes an instance of `StatsDConfiguration`. For the configuration's values, you will always need the StatsD server host name or IP address. Optionally, you can also change the port from the default (`8125`). You can also prepend a prefix to all stats. These values often come from configuration as the host name and/or prefix may vary between test and production environments.
+
+It is best practice to create a `StatsDPublisher` at application start and use it for the life of the application, instead of creating a new one for each usage.
 
 ## Setting up StatsD
 
@@ -87,6 +99,7 @@ An example of registering StatsD dependencies using `IServiceCollection` when us
 
 ```csharp
 // Simple
+// todo - use the SocketProtocol enum not the IpTransport class here
 services.AddSingleton<IStatsDTransport, IpTransport>();
 services.AddStatsD(Environment.GetEnvironmentVariable("MY_STATSD_ENDPOINT"));
 
@@ -97,11 +110,12 @@ services.AddStatsD(
     {
         var options = provider.GetRequiredService<MyOptions>().StatsD;
 
-        return new StatsDConfiguration()
+        return new StatsDConfiguration
         {
             Host = options.HostName,
             Port = options.Port,
             Prefix = options.Prefix,
+            SocketProtocol = SocketProtocol.Ip,
             OnError = ex => LogError(ex)
         };
     });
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,18 +11,30 @@
 
 ### TL;DR
 
-We use this within our components to publish [StatsD](http://github.com/etsy/statsd) metrics from .NET code. We've been using this in most of our things, since around 2013.
+We use this library within our components to publish [StatsD](http://github.com/etsy/statsd) metrics from .NET code to a server. We've been using this in most of our things, since around 2013.
+
+### Supported platforms
+
+`JustEat.Statsd` version 4 is built for these target frameworks: `net451` , `netstandard2.0`, `netcoreapp2.1`, `netcoreapp2.2`.
```
martincostello: Suggest making this a bulletpoint list.

README.md
```diff
@@ -11,18 +11,30 @@
...
 
 ### Features
 
-* StatsD metrics formatter
-* UDP client handling
+* Easy to use
+* Robust and proven
+* Tuned for high performance and low resource usage
+* Works well with modern .NET apps - .NET Core, .NET Standard 2,0 and `async ... await`
```
martincostello: Comma should be decimal place.

README.md
```diff
@@ -11,18 +11,30 @@
...
+* Supports standard statsd primitives: `Increment`, `Decrement`, `Timing` and `Gauge`. Supports sample rate.
```
martincostello: _statsd_ -> _StatsD_

martincostello: Suggest putting the sample rate on the next line as it isn't a primitive, or rephrasing slightly as it seems to not read well in the sentence.

README.md
```diff
@@ -11,18 +11,30 @@
...
+* Helpers to make it easy to time a `Func`, `Action` or a code block inside a `using` statement.
```
martincostello: Suggest changing to _"...a delegate, such as an `Action`, or a..."_ as `Func` isn't a real type.

README.md
```diff
@@ -11,18 +11,30 @@
...
+* Send stats over UDP or IP
+* Send stats to a server byt name or IP address
```
martincostello: _byt_ -> _by_

README.md
```diff
@@ -11,18 +11,30 @@
...
 
 #### Publishing statistics
 
-`IStatsDPublisher` is the interface that you will use in most circumstances. With this you can `Increment` or `Decrement` an event, and send values for a `Gauge` or `Timing`.
+`IStatsDPublisher` is the interface that you will use in most circumstances. With this you can `Increment` an event, and send values for a `Gauge` or `Timing`.
```
martincostello: This sentence implies there are other interfaces. Either we should give an example of an alternative, or rewrite the sentence to just say `IStatsDPublisher` and allude to no others.

README.md
```diff
@@ -11,18 +11,30 @@
...
 
-The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The constructor takes an instance of `StatsDConfiguration`. For the configuration's values, you will always need the StatsD server host name or IP address. Optionally, you can change the standard port (8125). You can also prepend a prefix to all stats. These values often come from configuration as the host name and/or prefix may vary between test and production environments.
+The concrete class that implements `IStatsDPublisher` is `StatsDPublisher`. The constructor takes an instance of `StatsDConfiguration`. For the configuration's values, you will always need the StatsD server host name or IP address. Optionally, you can also change the port from the default (`8125`). You can also prepend a prefix to all stats. These values often come from configuration as the host name and/or prefix may vary between test and production environments.
+
+It is best practice to create a `StatsDPublisher` at application start and use it for the life of the application, instead of creating a new one for each usage.
```
martincostello: Suggest _"...lifetime of the application, instead..."_

README.md
```diff
@@ -87,6 +99,7 @@ An example of registering StatsD dependencies using `IServiceCollection` when us
 
 ```csharp
 // Simple
+// todo - use the SocketProtocol enum not the IpTransport class here
```
martincostello: We should make an issue to track this, not TODO in the user-facing docs.

README.md
```diff
@@ -97,11 +110,12 @@ services.AddStatsD(
     {
         var options = provider.GetRequiredService<MyOptions>().StatsD;
 
-        return new StatsDConfiguration()
+        return new StatsDConfiguration
         {
             Host = options.HostName,
             Port = options.Port,
             Prefix = options.Prefix,
+            SocketProtocol = SocketProtocol.Ip,
```
martincostello: Suggest just changing this to `options.Protocol`, as otherwise the docs could be interpreted as saying _"you must set this to this in this scenario"_ as there is no other commentary here. Or, just omit it entirely as it seems more the sort of option that's for advanced usage that "normal config".

README.md
```diff
@@ -141,15 +154,16 @@ Bind<IStatsDPublisher>().To<StatsDPublisher>();
 | Host              | `string`                |                                | The host name or IP address of the StatsD server. There is no default, this must be set.                |
 | Port              | `int`                   | `8125`                         | The StatsD port.                                                                                        |
 | DnsLookupInterval | `TimeSpan?`             | `5 minutes`                    | Length of time to cache the host name to IP address lookup. Only used when "Host" contains a host name. |
-| Prefix            | `string`                | `string.Empty`                 | Prepend a prefix to all stats.                                                                          |
+| Prefix            | `string`                | `string.Empty`                 | Prepend a prefix to all stats.
+| SocketProtocol    | `enum`, one of `Udp`, `Ip`    | `Udp`                 | Type of socket to use when sending stats to the server.                                                                          |
```
martincostello: The type isn't `enum`, it's `SocketProtocol`.

README.md
```diff
@@ -141,15 +154,16 @@ Bind<IStatsDPublisher>().To<StatsDPublisher>();
...
 | OnError           | `Func<Exception, bool>` | `null`                         | Function to receive notification of any exceptions.                                                     |
 
 `OnError` is a function to receive notification of any errors that occur when trying to publish a metric. This function should return:
- *  **True** if the exception was handled and no further action is needed
- *  **False** if the exception should be thrown
 
-The default behaviour is to ignore the error.
+* **True** if the exception was handled and no further action is needed
```
martincostello: Suggest using `true` and `false` here, rather than bold.

martincostello:
You would use [this overload](https://github.com/justeat/JustEat.StatsD/blob/b74ac1d0a8586a4d0ddbeb38cc3d63ce13b2023e/src/JustEat.StatsD/StatsDServiceCollectionExtensions.cs#L83) to provide an entire config where the `IServiceProvider` is available.

An overload that specified the `StatsDConfiguration` as an input would require separate pre-bootstrap to get values from config as otherwise you're trying to configure it before DI registration is complete. If you _really_ wanted to do that yourself, there's nothing to stop your registering `StatsDConfiguration` that way with the service collection yourself.

README.md
AnthonySteele: AFIK, you _must_ set this in the AWS lambda scenario.

AnthonySteele:
Indeed, given that overload you can do
```
        public static IServiceCollection AddStatsD(this IServiceCollection services, StatsDConfiguration configuration)
        {
            return services.AddStatsD((s) => configuration);
        }
```

But as you pointed out that's not as useful, since you more likely want to populate the `StatsDConfiguration` with values from config etc, i.e. give it a func to call after the DI container is up and working.

README.md
martincostello: My mistake, I thought this was just the generic IoC example.

martincostello:
I'll look at doing a pass on the `///` docs later today.

PR #156, opened 2018-12-13
Comments review

### LINKED ISSUE #130: Documentation review

martincostello:
* Update XML documentation for the 4.0.0 release.
  * Make `StatsDUtf8Formatter` sealed.
  * Add parameter validation to `CachedEndpointSource`.
  * Rename constructor argument of `DisposableTimer` to match parameter names at call sites.

Resolves #130.

### DIFF
--- a/src/JustEat.StatsD/DisposableTimer.cs
+++ b/src/JustEat.StatsD/DisposableTimer.cs
@@ -12,16 +12,16 @@ namespace JustEat.StatsD
 
         public string StatName { get; set; }
 
-        public DisposableTimer(IStatsDPublisher publisher, string statName)
+        public DisposableTimer(IStatsDPublisher publisher, string bucket)
         {
             _publisher = publisher ?? throw new ArgumentNullException(nameof(publisher));
 
-            if (string.IsNullOrEmpty(statName))
+            if (string.IsNullOrEmpty(bucket))
             {
-                throw new ArgumentNullException(nameof(statName));
+                throw new ArgumentNullException(nameof(bucket));
             }
 
-            StatName = statName;
+            StatName = bucket;
             _stopwatch = Stopwatch.StartNew();
         }
 
--- a/src/JustEat.StatsD/DisposableTimer.cs
+++ b/src/JustEat.StatsD/DisposableTimer.cs
@@ -10,7 +10,7 @@ namespace JustEat.StatsD
         private IStatsDPublisher _publisher;
         private Stopwatch _stopwatch;
 
-        public string StatName { get; set; }
+        public string Bucket { get; set; }
 
         public DisposableTimer(IStatsDPublisher publisher, string bucket)
         {
@@ -21,7 +21,7 @@ namespace JustEat.StatsD
                 throw new ArgumentNullException(nameof(bucket));
             }
 
-            StatName = bucket;
+            Bucket = bucket;
             _stopwatch = Stopwatch.StartNew();
         }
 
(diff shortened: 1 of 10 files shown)


src/JustEat.StatsD/SocketProtocol.cs
```diff
@@ -1,9 +1,11 @@
+using System.Net.Sockets;
+
 namespace JustEat.StatsD
 {
     /// <summary>
-    /// The subset of ProtocolType that are supported by SocketTransport
-    /// UDP or IP.
-    /// UDP is the default, but IP transport is required for AWS Lambdas.
+    /// An emueration defining the subset of <see cref="ProtocolType"/> values that are supported by <see cref="SocketTransport"/>.
```
AnthonySteele: emueration -> enumeration

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -120,6 +138,9 @@ public static T Time<T>(this IStatsDPublisher publisher, string bucket, Func<T>
         /// <returns>
         /// The value from invoking <paramref name="func"/>.
         /// </returns>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="publisher"/> or <paramref name="bucket"/> is <see langword="null"/>.
+        /// </exception>
```
AnthonySteele: Should we do anything special if the `func` is null for these cases?

martincostello: I was pondering that - either we do nothing (as we did before) and let people trip over their `NullReferenceException`s, or I add null-checking to all these methods.

src/JustEat.StatsD/DisposableTimer.cs
```diff
@@ -12,16 +12,16 @@ internal sealed class DisposableTimer : IDisposableTimer
...
 
-            if (string.IsNullOrEmpty(statName))
+            if (string.IsNullOrEmpty(bucket))
             {
-                throw new ArgumentNullException(nameof(statName));
+                throw new ArgumentNullException(nameof(bucket));
             }
 
-            StatName = statName;
+            StatName = bucket;
```
AnthonySteele: Should we change the `public string StatName` as well?

src/JustEat.StatsD/TimerExtensions.cs
martincostello: You wouldn't really document `NullReferenceException` because that's just a "bug". Either we add the null-checking and document, or leave as-is.

I'd lean towards to the null-check even if it's adding an extra CPU instruction to each method.

src/JustEat.StatsD/DisposableTimer.cs
AnthonySteele: IMHO yes, for more consistent naming

src/JustEat.StatsD/TimerExtensions.cs
martincostello: Have added null checks in f8a61978d2b122d66b925918bab32d99d14e0f3e.

src/JustEat.StatsD/IDisposableTimer.cs
```diff
@@ -10,6 +10,6 @@ public interface IDisposableTimer : IDisposable
         /// <summary>
         /// Gets or sets the name of the StatsD bucket associated with the timer.
         /// </summary>
-        string StatName { get; set; }
+        string Bucket { get; set; }
```
AnthonySteele: Simpler as " Gets or sets the StatsD bucket..."

PR #158, opened 2018-12-18
version metadata for 4.0 beta 01

AnthonySteele:

Prepare for a beta of `4.0.0`. Final docs will be done when this beta has been used in 1 or more places.  

This PR generates `JustEat.StatsD.4.0.0-beta-01-build569.nupkg` [over here](https://ci.appveyor.com/project/justeattech/justeat-statsd/builds/21094058)

So I think, before we push to nuget, we would merge and tag?

### DIFF
--- a/version.props
+++ b/version.props
@@ -3,7 +3,7 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
     <VersionPrefix>4.0.0</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
+    <VersionSuffix>beta-01</VersionSuffix>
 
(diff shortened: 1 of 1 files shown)


martincostello:
I thought we were doing labeling through the GitHub UI to set the version so we don't need to do these commits, and then manually publishing those from AppVeyor if we liked the resulting builds?

AnthonySteele:
🤷‍♂️ IDK man, I'm asking becasue I'm not sure. we spend so much time automating it, it would be easier to do it by hand.

martincostello:
It is: https://github.com/justeat/JustEat.StatsD/blob/master/.github/CONTRIBUTING.md#releases

The only bit that is out of date is the auto-publishing, since you turned it off: #96

We _spent_ the time automating it, and now we don't need to churn the code base to increment version metadata that's then immediately out of date. We should only need to increment the `x.x.x` part of the number, not the prelease tags.

PR #167, opened 2019-02-17
Improve code coverage

martincostello:
* Add unit tests for uncovered members and code branches.
  * Remove code that was only exercised by tests and benchmarks and wasn't otherwise used.
  * If the bucket name is null for some reason, don't bother trying to send a stat.
  * Rename file to match type.


src/JustEat.StatsD/Buffered/BufferBasedStatsDPublisher.cs
```diff
@@ -43,7 +43,7 @@ public void Timing(long duration, double sampleRate, string bucket)
 
         private void SendMessage(double sampleRate, in StatsDMessage msg)
         {
-            bool shouldSendMessage = sampleRate >= DefaultSampleRate || sampleRate > Random.NextDouble();
+            bool shouldSendMessage = (sampleRate >= DefaultSampleRate || sampleRate > Random.NextDouble()) && msg.StatBucket != null;
```
AnthonySteele: Is this a place to use `! String.IsNullOrEmpty(msg.StatBucket)` ?

AnthonySteele:
Typically "Endpoint" not "EndPoint" : https://en.wikipedia.org/wiki/Endpoint

AnthonySteele:
Are there any changes that could be considered "breaking" here? Would like to avoid those now.

src/JustEat.StatsD/Buffered/BufferBasedStatsDPublisher.cs
martincostello: If we want to. I was using null to try and force the exception branch of the code but couldn’t, it was just pointlessly sending a metric, so just added this in instead.

martincostello:
There shouldn’t be any breaking changes here. All I did is delete an internal class and make something not send bad data.

PR #172, opened 2019-04-08
Allow IEndpointSource to return null Endpoint

### LINKED ISSUE #171: [Feature Request] Skip sending metrics if endpoint is not available
I'm using this library in an environment where everything in running in Docker containers with non-DNS based service discovery.  

To use this library, I have a custom implementation of `IEndPointSource` that hooks up to the service discovery framework and will always return an up to date `Endpoint` with the right IP address of Statsd server. So `SocketTransport` instance will always know how to talk to Statsd.

However there are cases where the Statsd server isn't running or is unavailable for an extended period of time. Now since metrics aren't critical to running an app, it should be ok to skip sending metrics until Statsd is back up and operational. However the problem is with UDP sockets on Linux, if the endpoint isn't running it will return a connection refused error (this behaviour is Linux only based on this issue https://github.com/dotnet/corefx/issues/31206) 

I was thinking if it's possible to allow `IEndPointSource` to return null from `GetEndpoint` to indicate the endpoint isn't running or not available. Then the `Send` method in `SocketTransport` can handle this condition by returning immediately instead of attempting to use a Socket to send metrics.

So this code block
```csharp
public void Send(in ArraySegment<byte> metric)
{
    if (metric.Array == null || metric.Count == 0)
    {
        return;
    }

    var pool = GetPool(_endpointSource.GetEndpoint());
    var socket = pool.PopOrCreate();
```
Would change to
```csharp
public void Send(in ArraySegment<byte> metric)
{
    if (metric.Array == null || metric.Count == 0)
    {
        return;
    }

    if (endpoint == null)
    {
        return;
    }

    var pool = GetPool(endpoint);
    var socket = pool.PopOrCreate();
```

The current behaviour doesn't cause any exceptions to the app since by default errors will be ignored, however the socket refused error usually adds a bit of delay for each failed send call and with the approach above this can be avoided.

yousifh:
Let IEndpointSource return a null Endpoint during sending metrics
to allow them to be skipped.

Resolves #171

### DIFF
--- a/tests/JustEat.StatsD.Tests/SocketTransportTests.cs
+++ b/tests/JustEat.StatsD.Tests/SocketTransportTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Net;
 using JustEat.StatsD.EndpointLookups;
+using Moq;
 using Shouldly;
 using Xunit;
 
@@ -69,6 +70,17 @@ namespace JustEat.StatsD
             using (var transport = new SocketTransport(LocalStatsEndpoint(), SocketProtocol.IP))
             {
                 transport.Send(new ArraySegment<byte>(Array.Empty<byte>()));
+            }
+        }
+
+        [Fact]
+        public static void SocketTransportIsNoopForNullEndpoint()
+        {
+            var endpointSource = Mock.Of<IEndPointSource>();
+
+            using (var transport = new SocketTransport(endpointSource, SocketProtocol.IP))
+            {
+                transport.Send("teststat:1|c");
             }
         }
 
--- a/src/JustEat.StatsD/SocketTransport.cs
+++ b/src/JustEat.StatsD/SocketTransport.cs
@@ -44,7 +44,13 @@ namespace JustEat.StatsD
                 return;
             }
 
-            var pool = GetPool(_endpointSource.GetEndpoint());
+            var endpoint = _endpointSource.GetEndpoint();
+            if (endpoint == null)
+            {
+                return;
+            }
+
+            var pool = GetPool(endpoint);
             var socket = pool.PopOrCreate();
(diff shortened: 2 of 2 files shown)


yousifh:
I'm a little confused by the code coverage results. I didn't expect files that I didn't modify to be impacted by this change. I haven't used Codecov before so I hope I'm not missing something obvious if there are some more tests that I need to add.

tests/JustEat.StatsD.Tests/SocketTransportTests.cs
```diff
@@ -72,6 +73,17 @@ public static void SocketTransportIsNoopForEmptyArray()
...
         }
 
+        [Fact]
+        public static void SocketTransportIsNoopForNullEndpoint()
+        {
+            var endpointSource = Mock.Of<IEndPointSource>();
+
+            using (var transport = new SocketTransport(endpointSource, SocketProtocol.IP))
+            {
+                transport.Send("teststat:1|c");
```
martincostello: So this test fails when your change is absent right?

martincostello:
There's some level of randomness in the tests that we haven't eliminated, so sometimes the coverage results vary for code unrelated to a change.

It looks to me that you've added the appropriate test case to handle this change.

tests/JustEat.StatsD.Tests/SocketTransportTests.cs
yousifh: Yes, without my change it would throw a null reference exception when the sockets pool is being created.

martincostello:
@yousifh Thanks for your contribution 👍 

I'll look at publishing 4.0.1 to NuGet.org with this change later this afternoon.

Issue #268, opened 2020-08-23
Support for tags/dimensions


gomete:
Hi all!

It would be amazing to allow tags. AFAIK, it is not part of the specification but it is supported by some of the major providers:

- https://docs.datadoghq.com/getting_started/tagging/
- https://www.datadoghq.com/blog/the-power-of-tagged-metrics/
- https://docs.splunk.com/Documentation/Splunk/8.0.5/Metrics/GetMetricsInStatsd
- https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-statsd.html#adding-dimensions-to-statsd-metrics
- https://www.influxdata.com/blog/getting-started-with-sending-statsd-metrics-to-telegraf-influxdb/#introducing-influx-statsd
- https://github.com/librato/statsd-librato-backend#tags

There are [several flavours to add tags based on the provider](https://github.com/prometheus/statsd_exporter).

There is already [a pull request that covers tagging](https://github.com/justeat/JustEat.StatsD/pull/229) but it does not support at least the 4 flavours depicted above.

martincostello:
Sorry for the incredibly long delay - tagging is now supported in JustEat.StatsD 5.0.0 which was just published to NuGet.org.

PR #229, opened 2020-03-06
tagging and gauge operations

gandarez:
* add support for tagging
* add support for gauge operations

Related to the issues [175](https://github.com/justeat/JustEat.StatsD/issues/175) and [176](https://github.com/justeat/JustEat.StatsD/issues/176).

PR #269, opened 2020-08-23
Tagging

gomete:
Support of tags or dimensions for the following providers:

- [DataDog](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics), [Splunk](https://docs.splunk.com/Documentation/Splunk/8.0.5/Metrics/GetMetricsInStatsd) and [CloudWatch](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-custom-metrics-statsd.html).
- [InfluxDB](https://www.influxdata.com/blog/getting-started-with-sending-statsd-metrics-to-telegraf-influxdb/#introducing-influx-statsd).
- [Librato](https://github.com/librato/statsd-librato-backend#tags).
- [SignalFX](https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-statsd.html#adding-dimensions-to-statsd-metrics).

It covers issue: #268 

It contains a breaking change. The interface `IStatsDPublisher` has been modified to include the tags as `Dictionary<string, string?>`. The old methods of this interface have been included as extension methods at `IStatsDPublisherExtensions`.
Therefore:
- Implementations of `IStatsDPublisher` will be broken after the change.
- Former usage of `IStatsDPublisher` and `StatsDPublisher` will work as it was in the previous version.

### After the change
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1016 (1909/November2018Update/19H2)
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.401
  [Host]     : .NET Core 3.1.7 (CoreCLR 4.700.20.36602, CoreFX 4.700.20.37001), X64 RyuJIT
  DefaultJob : .NET Core 3.1.7 (CoreCLR 4.700.20.36602, CoreFX 4.700.20.37001), X64 RyuJIT


```
|                                   Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------------- |---------:|--------:|--------:|------:|------:|------:|----------:|
|                              BufferBased | 248.0 ns | 4.58 ns | 4.29 ns |     - |     - |     - |         - |
|                  BufferBasedIgnoringTags | 272.0 ns | 3.61 ns | 3.01 ns |     - |     - |     - |         - |
|             BufferBasedIgnoringEmptyTags | 256.3 ns | 5.16 ns | 5.94 ns |     - |     - |     - |         - |
|  BufferBasedWithTagsFormatterAndNullTags | 248.9 ns | 4.86 ns | 5.59 ns |     - |     - |     - |         - |
|             BufferBasedWithTagsFormatter | 661.2 ns | 4.65 ns | 4.35 ns |     - |     - |     - |         - |
| BufferBasedWithTagsFormatterAndEmptyTags | 252.7 ns | 3.36 ns | 2.98 ns |     - |     - |     - |         - |

### Previous version
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1016 (1909/November2018Update/19H2)
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.401
  [Host]     : .NET Core 3.1.7 (CoreCLR 4.700.20.36602, CoreFX 4.700.20.37001), X64 RyuJIT
  DefaultJob : .NET Core 3.1.7 (CoreCLR 4.700.20.36602, CoreFX 4.700.20.37001), X64 RyuJIT


```
|      Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------ |---------:|--------:|--------:|------:|------:|------:|----------:|
| BufferBased | 226.8 ns | 4.48 ns | 3.97 ns |     - |     - |     - |         - |


src/JustEat.StatsD/Buffered/Tags/SignalFxFormatter.cs
```diff
@@ -0,0 +1,17 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    internal sealed class SignalFxFormatter : BaseTagsFormatter
```
martincostello, line 3: Similar comments as for the DataDog one.

src/JustEat.StatsD/Buffered/Tags/StatsDTagsFormatter.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace JustEat.StatsD.Buffered.Tags
+{
+    internal sealed class StatsDTagsFormatter : IStatsDTagsFormatter
+    {
+        private readonly IStatsDTagsFormatter _inner;
+
+        public StatsDTagsFormatter(TagsStyle tagsStyle)
+        {
+            _inner = tagsStyle switch
```
martincostello, line 12: As noted earlier in the review, let's expose this without an enum and require the user to configure a specific `IStatsDTagsFormatter` if they want to use tags.

src/JustEat.StatsD/Buffered/Tags/TagsLocation.cs
```diff
@@ -0,0 +1,8 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    internal enum TagsLocation
```
martincostello, line 3: I think we can just simplify this to a `bool` where it's used - anything different the consumer could just implement the interface themselves.

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -59,8 +105,14 @@ public static void Increment(this IStatsDPublisher publisher, long value, double
...
         /// <param name="value">The value to increment the counter(s) by.</param>
         /// <param name="sampleRate">The sample rate for the counter(s).</param>
+        /// <param name="tags">The tags to publish with the counter.</param>
         /// <param name="buckets">The bucket(s) to increment the counter(s) for.</param>
-        public static void Increment(this IStatsDPublisher publisher, long value, double sampleRate, params string[] buckets)
+        public static void Increment(
+            this IStatsDPublisher publisher,
+            long value,
+            double sampleRate,
+            IDictionary<string, string?>? tags = null,
```
martincostello, line 114: I'm not sure this parameter will work well being optional if it is followed by the non-optional `params` argument. I think we'd want to do this via overloads.

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -12,14 +12,50 @@ public static class IStatsDPublisherExtensions
...
+            publisher.Timing(duration, sampleRate, bucket, null);
+        }
+
         /// <summary>
         /// Publishes a counter for the specified bucket with a value of one (1).
         /// </summary>
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="bucket">The bucket to increment the counter for.</param>
-        public static void Increment(this IStatsDPublisher publisher, string bucket)
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 55: Nit: on this one (and all the other new parameters) for consistency with the others should be `tag(s)`.

```suggestion
        /// <param name="tags">The tag(s) to publish with the counter.</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -28,9 +64,14 @@ public static void Increment(this IStatsDPublisher publisher, string bucket)
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="value">The value to increment the counter by.</param>
         /// <param name="bucket">The bucket to increment the counter for.</param>
-        public static void Increment(this IStatsDPublisher publisher, long value, string bucket)
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 67:
```suggestion
        /// <param name="tags">The tag(s) to publish with the counter.</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -40,7 +81,12 @@ public static void Increment(this IStatsDPublisher publisher, long value, string
         /// <param name="value">The value to increment the counter(s) by.</param>
         /// <param name="sampleRate">The sample rate for the counter(s).</param>
         /// <param name="buckets">The bucket(s) to increment the counter(s) for.</param>
-        public static void Increment(this IStatsDPublisher publisher, long value, double sampleRate, IEnumerable<string> buckets)
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 84:
```suggestion
        /// <param name="tags">The tag(s) to publish with the counter(s).</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -59,8 +105,14 @@ public static void Increment(this IStatsDPublisher publisher, long value, double
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="value">The value to increment the counter(s) by.</param>
         /// <param name="sampleRate">The sample rate for the counter(s).</param>
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 108:
```suggestion
        /// <param name="tags">The tag(s) to publish with the counter.</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -78,9 +130,10 @@ public static void Increment(this IStatsDPublisher publisher, long value, double
         /// </summary>
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="bucket">The bucket to decrement the counter for.</param>
-        public static void Decrement(this IStatsDPublisher publisher, string bucket)
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 133:
```suggestion
        /// <param name="tags">The tag(s) to publish with the counter.</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -89,9 +142,14 @@ public static void Decrement(this IStatsDPublisher publisher, string bucket)
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="value">The value to decrement the counter by.</param>
         /// <param name="bucket">The bucket to decrement the counter for.</param>
-        public static void Decrement(this IStatsDPublisher publisher, long value, string bucket)
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 145:
```suggestion
        /// <param name="tags">The tag(s) to publish with the counter.</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -101,9 +159,15 @@ public static void Decrement(this IStatsDPublisher publisher, long value, string
         /// <param name="value">The value to decrement the counter by.</param>
         /// <param name="sampleRate">The sample rate for the counter.</param>
         /// <param name="bucket">The bucket to decrement the counter for.</param>
-        public static void Decrement(this IStatsDPublisher publisher, long value, double sampleRate, string bucket)
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 162:
```suggestion
        /// <param name="tags">The tag(s) to publish with the counter.</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -113,7 +177,13 @@ public static void Decrement(this IStatsDPublisher publisher, long value, double
         /// <param name="value">The value to decrement the counter(s) by.</param>
         /// <param name="sampleRate">The sample rate for the counter(s).</param>
         /// <param name="buckets">The bucket(s) to decrement the counter(s) for.</param>
-        public static void Decrement(this IStatsDPublisher publisher, long value, double sampleRate, IEnumerable<string> buckets)
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 180:
```suggestion
        /// <param name="tags">The tag((s) to publish with the counter(s).</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -134,8 +204,14 @@ public static void Decrement(this IStatsDPublisher publisher, long value, double
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="value">The value to decrement the counter(s) by.</param>
         /// <param name="sampleRate">The sample rate for the counter(s).</param>
+        /// <param name="tags">The tags to publish with the counter.</param>
```
martincostello, line 207:
```suggestion
        /// <param name="tags">The tag(s) to publish with the counter(s).</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -134,8 +204,14 @@ public static void Decrement(this IStatsDPublisher publisher, long value, double
...
         /// <param name="buckets">The bucket(s) to decrement the counter(s) for.</param>
-        public static void Decrement(this IStatsDPublisher publisher, long value, double sampleRate, params string[] buckets)
+        public static void Decrement(
+            this IStatsDPublisher publisher,
+            long value,
+            double sampleRate,
+            IDictionary<string, string?>? tags = null,
```
martincostello, line 213: Same here regarding the optional parameter and `params`.

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -156,9 +232,14 @@ public static void Decrement(this IStatsDPublisher publisher, long value, double
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="duration">The value to publish for the timer.</param>
         /// <param name="bucket">The bucket to publish the timer for.</param>
-        public static void Timing(this IStatsDPublisher publisher, TimeSpan duration, string bucket)
+        /// <param name="tags">The tags to publish with the timer.</param>
```
martincostello, line 235:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -168,9 +249,15 @@ public static void Timing(this IStatsDPublisher publisher, TimeSpan duration, st
         /// <param name="duration">The value to publish for the timer.</param>
         /// <param name="sampleRate">The sample rate for the timer.</param>
         /// <param name="bucket">The bucket to publish the timer for.</param>
-        public static void Timing(this IStatsDPublisher publisher, TimeSpan duration, double sampleRate, string bucket)
+        /// <param name="tags">The tags to publish with the timer.</param>
```
martincostello, line 252:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
```diff
@@ -179,9 +266,14 @@ public static void Timing(this IStatsDPublisher publisher, TimeSpan duration, do
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="duration">The value to publish for the timer.</param>
         /// <param name="bucket">The bucket to publish the timer for.</param>
-        public static void Timing(this IStatsDPublisher publisher, long duration, string bucket)
+        /// <param name="tags">The tags to publish with the timer.</param>
```
martincostello, line 269:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -51,6 +51,11 @@ public class StatsDConfiguration
         /// </summary>
         public string Prefix { get; set; } = string.Empty;
 
+        /// <summary>
+        /// Gets or sets the style for tagging. By default Disabled.
+        /// </summary>
+        public TagsStyle TagsStyle { get; set; } = TagsStyle.Disabled;
```
martincostello, line 57: As noted earlier, let's just have this be the interface for the formatter. It can have the default value of the no-op one (or even `null` and internally it would `?? new NoOp...()` at the point of assignment).

src/JustEat.StatsD/TagsStyle.cs
```diff
@@ -0,0 +1,35 @@
+namespace JustEat.StatsD
+{
+    /// <summary>
+    /// An enumeration defining the tagging formats supported.
+    /// <para />
+    /// Disabled is the default.
+    /// </summary>
+    public enum TagsStyle
```
martincostello, line 8: As noted earlier, let's remove the need for this enum so it's extensible without needing a whole new release for a new provider.

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -15,15 +16,19 @@ public static class TimerExtensions
         /// </summary>
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="bucket">The bucket to publish the timer for.</param>
+        /// <param name="tags">The tags to publish with the timer.</param>
```
martincostello, line 19:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -33,17 +38,22 @@ public static IDisposableTimer StartTimer(this IStatsDPublisher publisher, strin
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="bucket">The bucket to publish the timer for.</param>
         /// <param name="action">A delegate to a method whose invocation should be timed.</param>
+        /// <param name="tags">The tags to publish with the timer.</param>
```
martincostello, line 41:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -56,17 +66,22 @@ public static void Time(this IStatsDPublisher publisher, string bucket, Action a
...
```
martincostello, line 69:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -79,74 +94,88 @@ public static void Time(this IStatsDPublisher publisher, string bucket, Action<I
...
```
martincostello, line 97:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -79,74 +94,88 @@ public static void Time(this IStatsDPublisher publisher, string bucket, Action<I
...
 
-
         /// <summary>
         /// Starts a new timer for the specified bucket which is started when the specified delegate is invoked
         /// and awaited, and is stopped and published when the asynchronous delegate invocation completes.
         /// </summary>
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="bucket">The bucket to publish the timer for.</param>
         /// <param name="action">A delegate to a method whose invocation should be timed.</param>
+        /// <param name="tags">The tags to publish with the timer.</param>
```
martincostello, line 128:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -79,74 +94,88 @@ public static void Time(this IStatsDPublisher publisher, string bucket, Action<I
...
-            using (var timer = StartTimer(publisher, bucket))
+            using (var timer = StartTimer(publisher, bucket, tags))
             {
                 await action(timer).ConfigureAwait(false);
             }
         }
 
         /// <summary>
-        /// Starts a new timer for the specified bucket which is started when the specified delegate is invoked
+        /// Starts a new timer for the specified bucket and tags which is started when the specified delegate is invoked
```
martincostello, line 153:
```suggestion
        /// Starts a new timer for the specified bucket and any tag(s) which is started when the specified delegate is invoked
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -79,74 +94,88 @@ public static void Time(this IStatsDPublisher publisher, string bucket, Action<I
...
         /// and is stopped and published when the delegate invocation completes.
         /// </summary>
         /// <typeparam name="T">The type of the result of the delegate to invoke.</typeparam>
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="bucket">The bucket to publish the timer for.</param>
         /// <param name="func">A delegate to a method whose invocation should be timed and result returned.</param>
+        /// <param name="tags">The tags to publish with the timer.</param>
```
martincostello, line 160:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -79,74 +94,88 @@ public static void Time(this IStatsDPublisher publisher, string bucket, Action<I
...
         /// <returns>
         /// The value from invoking <paramref name="func"/>.
         /// </returns>
         /// <exception cref="ArgumentNullException">
-        /// <paramref name="publisher"/>, <paramref name="bucket"/> or <paramref name="func"/> is <see langword="null"/>.
+        /// <paramref name="publisher"/>, <paramref name="bucket"/>, <paramref name="tags"/> or <paramref name="func"/> is <see langword="null"/>.
```
martincostello, line 165: Does it throw if `tags` is `null`? The code I've seen so far in the review just seemed to skip tags if the value was null.

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -160,20 +189,25 @@ public static T Time<T>(this IStatsDPublisher publisher, string bucket, Func<T>
         /// <param name="publisher">The <see cref="IStatsDPublisher"/> to publish with.</param>
         /// <param name="bucket">The bucket to publish the timer for.</param>
         /// <param name="func">A delegate to a method whose invocation should be timed and result returned.</param>
+        /// <param name="tags">The tags to publish with the timer.</param>
```
martincostello, line 192:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -187,20 +221,25 @@ public static T Time<T>(this IStatsDPublisher publisher, string bucket, Func<IDi
...
```
martincostello, line 224:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

src/JustEat.StatsD/TimerExtensions.cs
```diff
@@ -214,20 +253,25 @@ public static async Task<T> Time<T>(this IStatsDPublisher publisher, string buck
...
```
martincostello, line 256:
```suggestion
        /// <param name="tags">The tag(s) to publish with the timer.</param>
```

martincostello: I'm not going to go back and edit the others now, but we should probably annotate this like the below to make clear that tags are not required.

```
The optional tag(s) to publish with the timer.
```

martincostello (commented):
Thanks for the PR @jgrivas - this is a great start!

I think we need to change this with regards to ease of extensibility, but otherwise this looks like a great addition. 😃

We should also compare it with the implementation from #229 to see the commonalities and differences.

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
```diff
@@ -1,18 +1,21 @@
...
 
 namespace JustEat.StatsD.Buffered
 {
     internal sealed class StatsDUtf8Formatter
     {
         private readonly byte[] _utf8Prefix;
+        private readonly IStatsDTagsFormatter _tagsFormatter;
 
-        public StatsDUtf8Formatter(string prefix)
+        public StatsDUtf8Formatter(string prefix, TagsStyle tagsStyle)
```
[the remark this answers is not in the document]
jgrivas: Sure. You're right. I was thinking only about making as simpler as possible to instantiate the StatsDPublisher and not in extensibility. I will change it today.

README.md
```diff
@@ -144,15 +148,16 @@ Bind<StatsDConfiguration>().ToConstant(statsDConfig>);
...
 | DnsLookupInterval | `TimeSpan?`             | `5 minutes`                    | Length of time to cache the host name to IP address lookup. Only used when "Host" contains a host name. |
-| Prefix            | `string`                | `string.Empty`                 | Prepend a prefix to all stats.
-| SocketProtocol    | `SocketProtocol`, one of `Udp`, `IP`| `Udp`              | Type of socket to use when sending stats to the server.                                                                          |
+| Prefix            | `string`                | `string.Empty`                 | Prepend a prefix to all stats.                                                                          |
+| SocketProtocol    | `SocketProtocol`, one of `Udp`, `IP`| `Udp`              | Type of socket to use when sending stats to the server.                                                 |
+| TagsStyle         | `TagsStyle`, one of `Disabled`, `DataDog`, `InfluxDb`, `Librato`, `SignalFx`| `Disabled` | Format used for tags fir the different providers.                       |
```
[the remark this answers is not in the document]
jgrivas: Hi Martin, thanks for the feedback! I will inject the IStatsDTagsFormatter instead.

README.md
```diff
@@ -162,7 +167,16 @@ Bind<IStatsDPublisher>().To<StatsDPublisher>().InSingletonScope();
...
 
-#### Example of using the interface
+#### Tagging support
+
+Tags or dimensions are not covered by the StatsD specification. Providers supporting tags have implemented their flavours:
+
+* [DataDog](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics) and [Splunk](https://docs.splunk.com/Documentation/Splunk/8.0.5/Metrics/GetMetricsInStatsd).
+* [InfluxDB](https://www.influxdata.com/blog/getting-started-with-sending-statsd-metrics-to-telegraf-influxdb/#introducing-influx-statsd).
+* [Librato](https://github.com/librato/statsd-librato-backend#tags).
+* [SignalFX](https://docs.signalfx.com/en/latest/integrations/agent/monitors/collectd-statsd.html#adding-dimensions-to-statsd-metrics).
```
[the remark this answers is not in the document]
jgrivas: Let me check the CloudWatch format and I will try to add it today as well. Thank you for the feedback!

src/JustEat.StatsD/TimerExtensions.cs
jgrivas: No, you're right. I forgot to remove it. I will do it.

src/JustEat.StatsD/Buffered/Tags/BaseTagsFormatter.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace JustEat.StatsD.Buffered.Tags
+{
+    internal abstract class BaseTagsFormatter : IStatsDTagsFormatter
```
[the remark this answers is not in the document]
jgrivas: I will add to the README.md file tagging section the way to extend it for other providers if you agree.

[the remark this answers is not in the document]
martincostello: Sure, that makes sense.

README.md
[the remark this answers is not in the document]
jgrivas: It seems that CloudWatch is supporting the same format than DataDof and Splunk: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-custom-metrics-statsd.html

src/JustEat.StatsD/TimerExtensions.cs
jgrivas: Done!

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
jgrivas: Changed with overloads.

src/JustEat.StatsD/IStatsDPublisherExtensions.cs
jgrivas: Changed with overloads.

src/JustEat.StatsD/Buffered/Tags/DataDogTagsFormatter.cs
```diff
@@ -0,0 +1,22 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    /// <summary>
+    /// Formats StatsD tags for DataDog, Splunk and CloudWatch.
```
martincostello, line 4: As this supports more than one provider, maybe it can have a more generic name?

src/JustEat.StatsD/Buffered/Tags/IStatsDTagsFormatter.cs
```diff
@@ -2,12 +2,32 @@
 
 namespace JustEat.StatsD.Buffered.Tags
 {
-    internal interface IStatsDTagsFormatter
```
martincostello: I don't think we'd want the ref struct to be in the public interface. Is there a way to expose this interface using strings in a simpler way, and then have an adapter internally like we have today that maps the user-specified one to the internal one that uses buffers?

src/JustEat.StatsD/Buffered/Tags/InfluxDbTagsFormatter.cs
```diff
@@ -0,0 +1,22 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    /// <summary>
+    /// Formats StatsD tags for InfluxDB.
+    /// Tags placed right after the bucket name with format: "," + tag1=value1,tag2,tag3=value
```
martincostello, line 5:
```suggestion
    /// Tags placed right after the bucket name with format: <code>"," + tag1=value1,tag2,tag3=value</code>
```

src/JustEat.StatsD/Buffered/Tags/DataDogTagsFormatter.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Tags placed at the end of the message with format: "|#" + tag1:value1,tag2,tag3:value
```
martincostello, line 5:
```suggestion
    /// Tags placed at the end of the message with format: <code>"|#" + tag1:value1,tag2,tag3:value</code>
```

src/JustEat.StatsD/Buffered/Tags/LibratoTagsFormatter.cs
```diff
@@ -0,0 +1,22 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    /// <summary>
+    /// Formats StatsD tags for Librato.
+    /// Tags placed right after the bucket name with format: "#" + tag1=value1,tag2,tag3=value
```
martincostello, line 5:
```suggestion
    /// Tags placed right after the bucket name with format: <code>"#" + tag1=value1,tag2,tag3=value</code>
```

src/JustEat.StatsD/Buffered/Tags/SignalFxTagsFormatter.cs
```diff
@@ -0,0 +1,23 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    /// <summary>
+    /// Formats StatsD tags for SignalFX.
+    /// Tags placed right after the bucket name with format: "[" + tag1=value1,tag2,tag3=value + "]"
```
martincostello, line 5:
```suggestion
    /// Tags placed right after the bucket name with format: <code>"[" + tag1=value1,tag2,tag3=value + "]"</code>
```

src/JustEat.StatsD/StatsDConfiguration.cs
```diff
@@ -52,9 +53,9 @@ public class StatsDConfiguration
         public string Prefix { get; set; } = string.Empty;
 
         /// <summary>
-        /// Gets or sets the style for tagging. By default Disabled.
+        /// Gets or sets the formatter for tags. By default NoOpTagsFormatter.
```
martincostello, line 56:
```suggestion
        /// Gets or sets the formatter for tags. By default an instance of <see cref="NoOpTagsFormatter"/> is used.
```

src/JustEat.StatsD/Buffered/Buffer.cs
```diff
@@ -2,15 +2,18 @@
...
 namespace JustEat.StatsD.Buffered
 {
-    internal ref struct Buffer
+    /// <summary>
+    /// A class representing the buffer where the StatsD message is written.
+    /// </summary>
+    public ref struct Buffer
     {
-        public Buffer(Span<byte> source)
+        internal Buffer(Span<byte> source)
```
martincostello, line 10: Would this type actually be usable by a custom implementation of the tags formatter that isn't within the StatsD assembly itself if all the members of the buffer are internal?

martincostello (commented):
I think the extensibility needs a few more tweaks, but otherwise I think this is most of the way there now 👍

README.md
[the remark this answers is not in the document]
jgrivas: Hi Martin, I have already injected the IStatsDTagsFormatter and include the supported out-of-the-box in the class SupportedTagsFormatter.

README.md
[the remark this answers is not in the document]
jgrivas: Included in the documentation and in the class SupportedTagsFormatter.

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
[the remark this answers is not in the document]
jgrivas: Hi Martin, I have already injected the IStatsDTagsFormatter and include the supported out-of-the-box in the class SupportedTagsFormatter.

src/JustEat.StatsD/Buffered/Tags/BaseTagsFormatter.cs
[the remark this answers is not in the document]
jgrivas: I have included how to extend it in the README.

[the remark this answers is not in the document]
jgrivas: Renamed to StatsDTagsFormatter.

src/JustEat.StatsD/Buffered/Tags/BaseTagsFormatter.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        private readonly byte[] _utf8Prefix;
+        private readonly byte[] _utf8Suffix;
+        private readonly TagsLocation _location;
```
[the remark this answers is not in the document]
jgrivas: I have changed it for a boolean value.

src/JustEat.StatsD/Buffered/Tags/DataDogFormatter.cs
```diff
@@ -0,0 +1,16 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    internal sealed class DataDogFormatter : BaseTagsFormatter
```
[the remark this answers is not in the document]
jgrivas: Renamed to DataDogTagsFormatter and made public.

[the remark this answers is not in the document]
jgrivas: Renamed.

src/JustEat.StatsD/Buffered/Tags/InfluxDbFormatter.cs
```diff
@@ -0,0 +1,16 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    internal sealed class InfluxDbFormatter : BaseTagsFormatter
```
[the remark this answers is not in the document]
jgrivas: Renamed to InfluxDbTagsFormatter and made public.

src/JustEat.StatsD/Buffered/Tags/LibratoFormatter.cs
```diff
@@ -0,0 +1,16 @@
+namespace JustEat.StatsD.Buffered.Tags
+{
+    internal sealed class LibratoFormatter : BaseTagsFormatter
```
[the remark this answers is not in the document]
jgrivas: Renamed to LibratoTagsFormatter and made public.

src/JustEat.StatsD/Buffered/Tags/SignalFxFormatter.cs
jgrivas: Renamed to SignalFxTagsFormatter and made public.

src/JustEat.StatsD/Buffered/Tags/StatsDTagsFormatter.cs
jgrivas: Done.

src/JustEat.StatsD/Buffered/Tags/TagsLocation.cs
jgrivas: I have changed it by the bool "areBucketNameTags".

src/JustEat.StatsD/StatsDConfiguration.cs
jgrivas: Done.

src/JustEat.StatsD/TagsStyle.cs
jgrivas: Done.

src/JustEat.StatsD/Buffered/Tags/IStatsDTagsFormatter.cs
jgrivas: Sure. I will change the `IStatsDTagsFormatter` to return a string and add it to the buffer internally.

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
+using JustEat.StatsD.Buffered.Tags;
+
+namespace JustEat.StatsD
+{
+    /// <summary>
+    /// Supported <see cref="IStatsDTagsFormatter"/> implementations for some of the major providers.
```
martincostello, line 6:
```suggestion
    /// A class that can be used to retrieve supported <see cref="IStatsDTagsFormatter"/> implementations for some of the major metric providers.
```

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public static class SupportedTagsFormatter
+    {
+        /// <summary>
+        /// Gets the AWS CloudWatch tags formatter.
```
martincostello, line 11: Changing _the_ to _a(n)_ as they're new objects on every access.

```suggestion
        /// Gets an AWS CloudWatch tags formatter.
```

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public static IStatsDTagsFormatter CloudWatch => new DataDogTagsFormatter();
+
+        /// <summary>
+        /// Gets the DataDog tags formatter.
```
martincostello, line 16:
```suggestion
        /// Gets a DataDog tags formatter.
```

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public static IStatsDTagsFormatter DataDog => new DataDogTagsFormatter();
+        
+        /// <summary>
+        /// Gets the InfluxDB tags formatter.
```
martincostello, line 21:
```suggestion
        /// Gets an InfluxDB tags formatter.
```

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public static IStatsDTagsFormatter InfluxDb => new InfluxDbTagsFormatter();
+        
+        /// <summary>
+        /// Gets the Librato tags formatter.
```
martincostello, line 26:
```suggestion
        /// Gets a Librato tags formatter.
```

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public static IStatsDTagsFormatter Librato => new LibratoTagsFormatter();
+        
+        /// <summary>
+        /// Gets the SignalFX dimensions formatter.
```
martincostello, line 31:
```suggestion
        /// Gets a SignalFX dimensions formatter.
```

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public static IStatsDTagsFormatter SignalFx => new SignalFxTagsFormatter();
+        
+        /// <summary>
+        /// Gets the Splunk tags formatter.
```
martincostello, line 36:
```suggestion
        /// Gets a Splunk tags formatter.
```

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
+using JustEat.StatsD.Buffered.Tags;
+
+namespace JustEat.StatsD
+{
+    /// <summary>
+    /// Supported <see cref="IStatsDTagsFormatter"/> implementations for some of the major providers.
+    /// </summary>
+    public static class SupportedTagsFormatter
```
martincostello, line 8: Nice - I like this 👍

README.md
```diff
@@ -169,12 +166,37 @@ The default behaviour is to ignore the error.
...
+
+The template class `StatsDTagsFormatter` can be inherited providing:
+
+* **prefix**: the string that will appear before the tag(s).
+* **suffix**: the string that will appear after the tag(s).
+* **areBucketNameTags**: a boolean indicating if the tag(s) are placed right after the bucket name (like it is supported by InfluxDB, Librato or SignalFX) or otherwise at the end of the StatsD message (like it is supported by AWS CloudWatch, DataDog or Splunk).
+* **tagsSeparator**: the string that will be placed between tags.
+* **keyValueSeparator**: the string that will be placed between the tag key and its value.
+
+A new implementation of `IStatsDTagsFormatter` can be written from scratch. For this purpose there are available extension methods to write in the `Buffer`.
```
martincostello, line 199: I guess this sentence can be dropped with the refactor to hide away the buffers?

README.md
```diff
@@ -157,7 +154,7 @@ Bind<IStatsDPublisher>().To<StatsDPublisher>().InSingletonScope();
 | DnsLookupInterval | `TimeSpan?`             | `5 minutes`                    | Length of time to cache the host name to IP address lookup. Only used when "Host" contains a host name. |
 | Prefix            | `string`                | `string.Empty`                 | Prepend a prefix to all stats.                                                                          |
 | SocketProtocol    | `SocketProtocol`, one of `Udp`, `IP`| `Udp`              | Type of socket to use when sending stats to the server.                                                 |
-| TagsStyle         | `TagsStyle`, one of `Disabled`, `DataDog`, `InfluxDb`, `Librato`, `SignalFx`| `Disabled` | Format used for tags for the different providers.                       |
+| TagsFormatter     | `IStatsDTagsFormatter`  | `NoOpTagsFormatter`            | Format used for tags for the different providers. Out-of-the-box formatter at `SupportedTagsFormatter`. |
```
martincostello, line 157:
```suggestion
| TagsFormatter     | `IStatsDTagsFormatter`  | `NoOpTagsFormatter`            | Format used for tags for the different providers. Out-of-the-box formatters can be accessed using the `SupportedTagsFormatter` class. |
```

src/JustEat.StatsD/Buffered/Tags/DataDogTagsFormatter.cs
jgrivas: TrailingTagsFormatter for example?

martincostello: That sounds reasonable.

src/JustEat.StatsD/Buffered/Buffer.cs
jgrivas: I have modified the interface IStatsDPublisher to avoid exposing the Buffer struct. I will review the performance of the different providers but I think that the extension method TryWriteUtf8String has to manage null string or empty strings.

martincostello (commented):
This is looking really good now 😃

I'll re-review this as a whole tomorrow after all the changes just to give it another check-over, but otherwise I think it's just the performance results left to see what the difference is (mostly just for the no-op case for regression).

src/JustEat.StatsD/Buffered/Tags/IStatsDTagsFormatter.cs
```diff
@@ -15,19 +20,10 @@ public interface IStatsDTagsFormatter
...
-        /// <summary>
-        /// Writes to the buffer the tag(s) if the are included and they are placed at the end of the StatsD message.
+        /// Calculates the tag(s) formatted to be included in the StatsD message.
         /// </summary>
-        /// <param name="buffer">The buffer where the tag(s) should be written.</param>
         /// <param name="tags">The tag(s) included.</param>
-        /// <returns>A value indicating whether the operation was performed successfully.</returns>
-        bool TryWriteSuffixTagsIfNeeded(ref Buffer buffer, in IDictionary<string, string?>? tags);
+        /// <returns>The tag(s) formatted.</returns>
+        string GetFormattedTags(in IDictionary<string, string?>? tags);
```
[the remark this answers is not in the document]
jgrivas: Done!

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
```diff
@@ -24,34 +27,37 @@ public int GetMaxBufferSize(in StatsDMessage msg)
...
+               && TryWriteSampleRateIfNeeded(ref buffer, sampleRate)
+               && _tagsFormatter.TryWriteSuffixTagsIfNeeded(ref buffer, msg.Tags);
 
             written = isFormattingSuccessful ? buffer.Written : 0;
             return isFormattingSuccessful;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private bool TryWriteBucketNameWithColon(ref Buffer buffer, string statBucket)
+        private bool TryWriteBucketNameWithColon(ref Buffer buffer, in StatsDMessage msg)
```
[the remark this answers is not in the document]
jgrivas: Hi Martin,

I have performed some performance tests with BenchmarkDotNet and the statsd Docker image (<https://hub.docker.com/r/statsd/statsd>).

I downgrade BenchmarkDotNet version to 0.12.0 because they have an issue with the latest lang version project property in the latest version 0.12.1 (https://github.com/dotnet/BenchmarkDotNet/issues/1452).

This is the master branch execution:
image

With the tagging support, there is some performance loss but the most worrying part is the allocation and performance loss when tags are written:
image

src/JustEat.StatsD/Buffered/Tags/StatsDTagsFormatter.cs
```diff
@@ -45,27 +45,27 @@ public virtual int GetTagsBufferSize(in IDictionary<string, string?>? tags)
             }
 
             return _prefix.Length
-                + Encoding.UTF8.GetByteCount(GetFormattedTags(tags!))
+                + Encoding.UTF8.GetByteCount(this.FormatTags(tags!))
```
martincostello:
```suggestion
                + Encoding.UTF8.GetByteCount(FormatTags(tags!))
```

src/JustEat.StatsD/Buffered/Tags/StatsDTagsFormatter.cs
```diff
@@ -45,27 +45,27 @@ public virtual int GetTagsBufferSize(in IDictionary<string, string?>? tags)
...
         }
         
         /// <inheritdoc />
-        public virtual string GetFormattedTags(in IDictionary<string, string?>? tags)
+        public virtual string FormatTags(in IDictionary<string, string?>? tags)
         {
             if (AreTagsPresent(tags))
             {
-                return _prefix + GetFormattedTags(tags!) + _suffix;
+                return _prefix + this.FormatTags(tags!) + _suffix;
```
martincostello:
```suggestion
                return _prefix + FormatTags(tags!) + _suffix;
```

src/JustEat.StatsD/Buffered/Tags/StatsDTagsFormatter.cs
```diff
@@ -45,27 +45,27 @@ public virtual int GetTagsBufferSize(in IDictionary<string, string?>? tags)
...
             }
 
             return string.Empty;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private string GetFormattedTags(IDictionary<string, string?> tags) =>
-            string.Join(_tagsSeparator,tags.Select(tag => GetFormattedTag(tag)));
+        private string FormatTags(IDictionary<string, string?> tags) =>
+            string.Join(_tagsSeparator,tags.Select(tag => this.FormatTags(tag)));
```
martincostello: I think you can save the Lambda here and just point directly to the method?

```suggestion
            string.Join(_tagsSeparator, tags.Select(FormatTags));
```

tests/Benchmark/Benchmark.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="$(MicrosoftNETFrameworkReferenceAssembliesVersion)" PrivateAssets="All" />
-    <PackageReference Include="BenchmarkDotNet" Version="0.12.1" />
+    <PackageReference Include="BenchmarkDotNet" Version="0.12.0" />
```
martincostello, line 10: How come you've downgraded it?

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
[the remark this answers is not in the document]
martincostello: Sorry, I didn't see this because it was collapsed 😄

Yeah I'm aware of that bug, I fixed it!

Rather than downgrade, can you apply this work around instead?

https://github.com/martincostello/api/blob/f1822671d833e57da83121af17bab92953de6176/tests/API.Benchmarks/API.Benchmarks.csproj#L10-L11

Would you be able to paste the Markdown from the benchmarks into the main body of the PR so they don't get lost?

Then they'll be a bit easier to consume and triage than a screenshot 😃

src/JustEat.StatsD/PublicAPI.Shipped.txt
```diff
@@ -1,4 +1,19 @@
 #nullable enable
+JustEat.StatsD.Buffered.Tags.InfluxDbTagsFormatter
```
martincostello, line 2: We should probably move these out of the `Buffered` namespace and into either the root namespace, or a new one called something like `TagFormatters`?

src/JustEat.StatsD/SupportedTagsFormatter.cs
```diff
@@ -0,0 +1,40 @@
+using JustEat.StatsD.Buffered.Tags;
+
+namespace JustEat.StatsD
+{
+    /// <summary>
+    /// A class that can be used to retrieve supported <see cref="IStatsDTagsFormatter"/> implementations for some of the major metric providers.
+    /// </summary>
+    public static class SupportedTagsFormatter
```
martincostello, line 8: Let's rename this to `SupportedTagsFormatters` or `TagFormatters`.

src/JustEat.StatsD/Buffered/Tags/StatsDTagsFormatter.cs
jgrivas: Done!

src/JustEat.StatsD/Buffered/Tags/StatsDTagsFormatter.cs
jgrivas: Done!

src/JustEat.StatsD/Buffered/Tags/StatsDTagsFormatter.cs
jgrivas: Done!

src/JustEat.StatsD/PublicAPI.Shipped.txt
jgrivas: I have moved them to the folder `TagsFormatters` matching the namespace `JustEat.StatsD.TagsFormatters`.

src/JustEat.StatsD/SupportedTagsFormatter.cs
jgrivas: I have renamed it to `TagsFormatter` using the singular form to avoid the collision with the namespace `JustEat.StatsD.TagsFormatters`.

src/JustEat.StatsD/Buffered/StatsDUtf8Formatter.cs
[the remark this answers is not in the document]
jgrivas: I have applied the workaround and the results are included in the PR body. I didn't know that there was this report in markdown. Good to know!

I didn't see either that you were the fixer :D.

martincostello:
This is looking pretty much there now. The performance penalty compared to previous versions seems negligible for the case where there's no tags, _maybe_ there's some scope to see if the performance/allocations of the tags version can be improved a little but otherwise this looks great.

I've asked a couple of other Just Eat developers to take a look at this, but otherwise I'm happy with this.

I'll also compare this to #229 to see if there's anything that approach has that this lacks.

jgrivas:
> This is looking pretty much there now. The performance penalty compared to previous versions seems negligible for the case where there's no tags, _maybe_ there's some scope to see if the performance/allocations of the tags version can be improved a little but otherwise this looks great.
> 
> I've asked a couple of other Just Eat developers to take a look at this, but otherwise I'm happy with this.
> 
> I'll also compare this to #229 to see if there's anything that approach has that this lacks.

I have modified the `StatsDTagsFormatter` to follow the same approach followed at `BufferBasedStatsDPublisher` but with a `Span<char>`. It removed the allocation into the heap and reduces the time from ~1400 ns to 700-800 ns. But:
- I changed the interface `IStatsDTagsFormatter` to return a `ReadOnlySpan<char>` instead of a `string` to avoid the allocation of the string.
- I have changed the interface `IStatsDPublisher` to receive an array to tuples `(string, string?)[]?` instead of a dictionary `IDictionary<string, string?>?` to avoid the allocation in the iteration.

If you agree with those changes, I will clean up a little bit the changes and upload it in a single commit so it can be reverted easily.

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1016 (1909/November2018Update/19H2)
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.401
  [Host]     : .NET Core 3.1.7 (CoreCLR 4.700.20.36602, CoreFX 4.700.20.37001), X64 RyuJIT
  DefaultJob : .NET Core 3.1.7 (CoreCLR 4.700.20.36602, CoreFX 4.700.20.37001), X64 RyuJIT


```
|                       Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| BufferBasedWithTagsFormatter | 711.0 ns | 11.65 ns | 14.74 ns |     - |     - |     - |         - |

martincostello:
>I have changed the interface IStatsDPublisher to receive an array to tuples (string, string?)[]? instead of a dictionary IDictionary<string, string?>? to avoid the allocation in the iteration.

I think using the `(string, string?)` syntax would cause issues when called as a library outside of the solution. What about using `KeyValuePair<string, string?>` instead?

If there's no performance different, `IEnumerable<>?` would be preferable to `[]?` as well I think, as it allows interop with more types (e.g. dictionaries, lists) rather than forcing the caller to use an array.

jgrivas:
> I think using the `(string, string?)` syntax would cause issues when called as a library outside of the solution. What about using `KeyValuePair<string, string?>` instead?

Sure. Not problem. There is no performance penalty. The syntax is little more verbose but that's all.

> If there's no performance difference, `IEnumerable<>?` would be preferable to `[]?` as well I think, as it allows interop with more types (e.g. dictionaries, lists) rather than forcing the caller to use an array.

It has the same problem than the `IDictionary<,>` (which implements IEnumerable<>). There is a little allocation of 192 B while iterating it. The mean time increases from 700-800 ns to 1000-1100 ns.

slang25:
This PR is looking great, I'm glad you took the time to try and make it 0 allocations as doing it later would result in breaking changes.

I think the dictionary needs to be passed in with a concrete type (like `Dictionary<string,string?>?`), going by this blog post: https://www.nimaara.com/beware-of-the-idictionary-tkey-tvalue/#howaboutilistticollectiontienumerabletvslistt

jgrivas:
> This PR is looking great, I'm glad you took the time to try and make it 0 allocations as doing it later would result in breaking changes.
> 
> I think the dictionary needs to be passed in with a concrete type (like `Dictionary<string,string?>?`, or the array you have now), going by this blog post: https://www.nimaara.com/beware-of-the-idictionary-tkey-tvalue/#howaboutilistticollectiontienumerabletvslistt

Yes! With `Dictionary<,>` there is no allocation and the API looks better than the `KeyValuePair<,>[]`. Initialization in the user side is easier and less verbose. And is the same scenario regarding the interop with different types.

I will clean the `Dictionary<,>` version and upload it.

Thanks for the feedback!

jgrivas:
I think there is still an allocation for .NET 4.6.1 and .NET Standard 2.0 while encoding the chars to UTF8 at `BufferExtensions` (line 65) for .NET Standard 2.1 and .NET Core 2.1+, I ran the benchmark and it seems that there are no allocation now. I have edited the results for .NET Core 3.1 benchmark in the PR body.

src/JustEat.StatsD/Buffered/BufferExtensions.cs
```diff
@@ -52,9 +52,39 @@ public static bool TryWriteUtf8String(this ref Buffer src, string str)
...
+            var bytes = Encoding.UTF8.GetBytes(chars.ToArray());
+            return src.TryWrite(bytes);
+#else
+            int written = 0;
+            try
+            {
+                written = Encoding.UTF8.GetBytes(chars, src.Tail);
+            }
+#pragma warning disable CA1031
+            catch (ArgumentException)
```
martincostello, line 74: I _think_ having a catch clause in a method breaks the ability to inline this method (maybe I'm thinking of `throw` rather than `catch`). It would be better to test the condition that causes the exception instead if possible, otherwise you can probable delete the attribute from this method and the one above.

src/JustEat.StatsD/TagsFormatters/StatsDTagsFormatter.cs
```diff
@@ -11,10 +12,20 @@ namespace JustEat.StatsD.TagsFormatters
...
-        private readonly string _suffix;
-        private readonly string _tagsSeparator;
-        private readonly string _keyValueSeparator;
+        private const int SafeUdpPacketSize = 512;
+
+        [ThreadStatic]
+        private static char[]? _buffer;
+        private static char[] Buffer => _buffer ??= new char[SafeUdpPacketSize];
+
+        private readonly char[] _prefix;
```
martincostello, line 21: Would `Span<char>` work for these? Then it could just use the strings on the config as-is rather than needing to call `ToArray()`?

src/JustEat.StatsD/TagsFormatters/StatsDTagsFormatter.cs
```diff
@@ -27,53 +38,135 @@ protected StatsDTagsFormatter(StatsDTagsFormatterConfiguration configuration)
...
+            if (!TryFormatTags(destination, tags, out int written))
             {
-                return _prefix + FormatTags(tags!) + _suffix;
+                var newSize = GetTagsBufferSize(tags);
+                _buffer = new char[newSize];
+                destination = Buffer;
+
+                if (!TryFormatTags(destination, tags, out written))
+                {
+                    throw new Exception("Failed to format tags.");
```
martincostello, line 86: Maybe `FormatException` instead?

src/JustEat.StatsD/TagsFormatters/StatsDTagsFormatter.cs
```diff
@@ -27,53 +38,135 @@ protected StatsDTagsFormatter(StatsDTagsFormatterConfiguration configuration)
...
+
+            return isFormattingSuccessful;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private string FormatTags(IDictionary<string, string?> tags) =>
-            string.Join(_tagsSeparator,tags.Select(FormatTags));
+        private bool TryWriteTags(ref Buffer<char> src, in Dictionary<string, string?> tags)
+        {
+            var enumerator = tags.GetEnumerator();
```
martincostello, line 109: Do you need to dispose of the enumerator?

tests/JustEat.StatsD.Tests/StatsDPublisherTests.cs
```diff
@@ -90,9 +90,9 @@ public static void Metrics_Not_Sent_If_Array_Is_Null_Or_Empty()
             var config = new StatsDConfiguration();
             var anyValidTags = new Dictionary<string, string?>
             {
-                ["foo"] = "bar",
```
martincostello: I prefer the new dictionary syntax as it's a bit terser.

tests/Benchmark/Utf8FormatterBenchmark.cs
```diff
@@ -12,11 +12,11 @@ public class Utf8FormatterBenchmark
         private static readonly StatsDUtf8Formatter FormatterBuffer = new StatsDUtf8Formatter("hello.world", new NoOpTagsFormatter());
         private static readonly StatsDUtf8Formatter FormatterWithTagsBuffer = new StatsDUtf8Formatter("hello.world", TagsFormatter.DataDog);
 
-        private static readonly IDictionary<string, string?> EmptyTags = new Dictionary<string, string?>();
-        private static readonly IDictionary<string, string?> AnyValidTags = new Dictionary<string, string?>
+        private static readonly Dictionary<string, string?> EmptyTags = new Dictionary<string, string?>();
+        private static readonly Dictionary<string, string?> AnyValidTags = new Dictionary<string, string?>
         {
-            ["key"] = "value",
```
martincostello: Why change the syntax style?

martincostello:
> I think there is still an allocation for .NET 4.6.1 and .NET Standard 2.0 while encoding the chars to UTF8 at `BufferExtensions` (line 65) for .NET Standard 2.1 and .NET Core 2.1+, I ran the benchmark and it seems that there are no allocation now. I have edited the results for .NET Core 3.1 benchmark in the PR body.

I think as long as "latest" is all good, then it isn't really a problem if 4.6.1 and NS 2.0 have a few allocations from their lack of full Span support. The chances are anyone keeping up to date with package versions who wants this version with tags on it would probably be running .NET 2.1+ anyway.

src/JustEat.StatsD/TagsFormatters/StatsDTagsFormatter.cs
jgrivas: I think that the Span<char> only can be field inside a struct, not in a class because it cannot go to the heap.

src/JustEat.StatsD/Buffered/BufferExtensions.cs
jgrivas: Yes, you are right. It has no effect.

src/JustEat.StatsD/TagsFormatters/StatsDTagsFormatter.cs
```diff
@@ -27,53 +38,135 @@ protected StatsDTagsFormatter(StatsDTagsFormatterConfiguration configuration)
...
+            while (isFormattingSuccessful && enumerator.MoveNext())
+            {
+                isFormattingSuccessful = TryWriteTag(ref src, enumerator.Current)
+                    && TryWriteTagsSeparator(ref src, index, tags);
+
+                index++;
+            }
+
+            return isFormattingSuccessful;
+        }
```
slang25, line 121:
```csharp
        private bool TryWriteTags(ref Buffer<char> src, in Dictionary<string, string?> tags)
        {
            var index = 0;
            foreach (var tag in tags)
            {
                var isFormattingSuccessful = TryWriteTag(ref src, tag)
                                             && TryWriteTagsSeparator(ref src, index++, tags);

                if (!isFormattingSuccessful)
                {
                    return false;
                }
            }

            return true;
        }
```

src/JustEat.StatsD/Buffered/BufferExtensions.cs
jgrivas: Done!

src/JustEat.StatsD/TagsFormatters/StatsDTagsFormatter.cs
jgrivas: Done!

src/JustEat.StatsD/TagsFormatters/StatsDTagsFormatter.cs
jgrivas: Done!

tests/JustEat.StatsD.Tests/StatsDPublisherTests.cs
jgrivas: Done!

tests/Benchmark/Utf8FormatterBenchmark.cs
jgrivas: Don't know. I have changed from to KeyValuePair<string, string?>[], array of tuples and I didn't notice that I have changed at the end. Reverted!

src/JustEat.StatsD/TagsFormatters/StatsDTagsFormatter.cs
```diff
@@ -0,0 +1,172 @@
...
```
slang25, line 121:
```suggestion
        private bool TryWriteTags(ref Buffer<char> src, in Dictionary<string, string?> tags)
        {
            var index = 0;
            foreach (var tag in tags)
            {
                var isFormattingSuccessful = TryWriteTag(ref src, tag)
                                             && TryWriteTagsSeparator(ref src, index++, tags);
                if (!isFormattingSuccessful)
                {
                    return false;
                }
            }
            return true;
        }
```

jgrivas: Done! I have renamed the `Buffer` variable from `src` to `buffer`.

jgrivas:
Hi @martincostello, any plan to merge these changes?

martincostello:
Hey @jgrivas - sorry for the delays on this. I'll look at merging this in a week or so once we deploy a minor release this week.

martincostello:
As part of this PR could you update the versions to `5.0.0.0` and `5.0.0` please?

https://github.com/justeat/JustEat.StatsD/blob/ebe0af4f4b0bb67a278fedb38a2e82a28d2c50e6/version.props#L3-L4

jgrivas:
I have already increased the major version @martincostello.

martincostello (approved):

martincostello:
This looks good to me now, but I've asked @slang25 to sign it off as well.

I'll be back in the "office" next Monday, so I can hopefully look into merging this and sorting out a release for it next week.

martincostello:
Sorry I haven't had the time to look into this this week, but will make an effort to look at it next week 😞

martincostello:
I don't seem to be able to get GitHub to accept my attempts to resolve the merge conflicts. If you can sort them out, then I'll merge this PR.

gomete:
> I don't seem to be able to get GitHub to accept my attempts to resolve the merge conflicts. If you can sort them out, then I'll merge this PR.

Hi @martincostello! Sure, I will try to do it this week. Sorry I missed this notification.

jgrivas:
I don't know what happen with the fork but it seems that now my repo is not forked from the JustEat repository. There is only two repos. Your's and the one I created with a new account...

image

jgrivas:
I resolve the conflicts and new commits are not refreshed in this branch. I can remove my repository keeping my branch locally, create again the fork from this repository and push the branch again. And create the PR again.

jgpayvision:
I have created this PR #312 from a different account doing again the fork pushing the same changes and it seems that there are no conflicts now. We can continue with that if you agree.

martincostello:
OK, but I'll need to re-review it in the new PR so will take me longer as I'll need to find the time to do that at some point, hopefully this week.

martincostello:
Superseded by #312.

PR #374, opened 2022-02-17
Update to .NET 7

martincostello:
Update to .NET 7.


martincostello:
Nerd sniping @slang25 as to how to fix the compiler warnings as I don't really understand the new `scoped` feature 😅

PR #416, opened 2022-07-21
Move tags to IStatsDPublisherWithTags

martincostello:
- Move tag-related methods to the new `IStatsDPublisherWithTags` interface and make `IStatsDPublisher` match what it previously did in version [4.2.0](https://github.com/justeat/JustEat.StatsD/blob/v4.2.0/src/JustEat.StatsD/IStatsDPublisher.cs).
- Change version from `5.0.0` to `4.3.0` now that the binary-breaking changes (should) have been removed.


slang25 (approved):
Yep, looks good 👍

PR #585, opened 2023-11-15
Support AoT

martincostello:
Mark assembly as AoT friendly.

Thought I'd turn it on and see, and it Just Worked ™️.


slang25 (approved):
Sounds like a plan

PR #588, opened 2023-12-04
Bump version back to 5.0.0

martincostello:
We dropped several TFMs since the last release, so we still need to bump the major version even with the changes in #416.

```diff
- <TargetFrameworks>net461;netstandard2.0;netstandard2.1;netcoreapp2.1;net5.0</TargetFrameworks>
+ <TargetFrameworks>net462;netstandard2.0;net6.0</TargetFrameworks>
```

Alternatively, if we still want to release as 4.3.0 then do a different PR to put them back.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ We use this library within our components to publish [StatsD](http://github.com/
 
 ### Supported platforms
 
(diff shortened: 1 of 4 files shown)


slang25 (approved):
Talked about it and it makes sense to bump the version

PR #818, opened 2025-01-15
Add GraphiteDB tags formatter

tuyndoan:
# Add GraphiteDB Tags Formatter

## Description
Add support for GraphiteDB tags formatting in StatsD metrics. The new formatter allows tags to be placed right after the bucket name using GraphiteDB's specific format: `;tag1=value1;tag2;tag3=value3`.

## Changes
- Add new `GraphiteDbTagsFormatter` class implementing StatsD tags formatting for GraphiteDB
- Add corresponding unit tests to validate the formatter behavior
- The formatter uses the following configuration:
  - Prefix: ";"
  - No suffix
  - Tags are placed immediately after metric name (non-trailing)
  - Tags separator: ";"
  - Key-value separator: "="

## Example Outputs
```text
# Counter with tags
prefix.bucket;foo=bar;empty;lorem=ipsum:128|c

# Timing with tags and sampling rate
prefix.bucket;foo=bar;empty;lorem=ipsum:128|ms|@0.5

# Gauge with floating point value
prefix.bucket;foo=bar;empty;lorem=ipsum:128.5|g

### DIFF
--- a/src/JustEat.StatsD/PublicAPI.Shipped.txt
+++ b/src/JustEat.StatsD/PublicAPI.Shipped.txt
@@ -118,6 +118,7 @@ static JustEat.StatsD.StatsDServiceCollectionExtensions.AddStatsD(this Microsoft
 static JustEat.StatsD.TagsFormatter.CloudWatch.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
 static JustEat.StatsD.TagsFormatter.DataDog.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
 static JustEat.StatsD.TagsFormatter.InfluxDb.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
+static JustEat.StatsD.TagsFormatter.GraphiteDb.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
 static JustEat.StatsD.TagsFormatter.Librato.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
 static JustEat.StatsD.TagsFormatter.SignalFx.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
 static JustEat.StatsD.TagsFormatter.Splunk.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/JustEat.StatsD/PublicAPI.Shipped.txt [hunk 1]
```diff
@@ -118,6 +118,7 @@ static JustEat.StatsD.StatsDServiceCollectionExtensions.AddStatsD(this Microsoft
 static JustEat.StatsD.TagsFormatter.CloudWatch.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
 static JustEat.StatsD.TagsFormatter.DataDog.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
 static JustEat.StatsD.TagsFormatter.InfluxDb.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
+static JustEat.StatsD.TagsFormatter.GraphiteDb.get -> JustEat.StatsD.TagsFormatters.IStatsDTagsFormatter!
```
martincostello, line 121: It's fine to leave it here, but typically when adding new APIs they go into the `Unshipped` file and then get "promoted" to this file after release.

The main reason for that is that the analyser is stricter on things added to the Unshipped file for if, for example, a new overload would create ambiguity and potentially be a breaking change.

src/JustEat.StatsD/TagsFormatter.cs
```diff
@@ -17,6 +17,11 @@ public static class TagsFormatter
     /// </summary>
     public static IStatsDTagsFormatter DataDog => new TrailingTagsFormatter();
 
+    /// <summary>
+    /// Gets a GraphiteDB tags formatter.
+    /// </summary>
+    public static IStatsDTagsFormatter GraphiteDb => new GraphiteDbTagsFormatter();
```
martincostello, line 23: This highlights that we should have probably made these singletons via `{ get; } = ...` rather than creating a new instance on every access.

I'll fix that in a separate PR after this is merged.

martincostello (commented):
Could you bump the version to 5.1.0 please?

https://github.com/justeattakeaway/JustEat.StatsD/blob/361a3c9534f592460c92d9610ca03e02a6e7a460/version.props#L4

Otherwise this looks good - thanks for the contribution!

martincostello (approved):

martincostello:
Thanks again for your contribution @tuyndoan - your changes are available in [JustEat.StatsD v5.1.0](https://www.nuget.org/packages/JustEat.StatsD/5.1.0) 📦 which is now available from NuGet.org 🙇

